PR #15, opened 2018-01-15
Allow (and ignore) booleans as children of components

### LINKED ISSUE #14: Roact should reify booleans as nil
I was trying to conditionally render Roact components like this:

```condition and Roact.createElement(...)```

This throws an error:
```
18:06:27.365 - ReplicatedStorage.Roact.Reconciler:238: Cannot reify invalid Roact element "false"
18:06:27.367 - Stack Begin
18:06:27.369 - Script 'ReplicatedStorage.Roact.Reconciler', Line 238 - field _reifyInternal
18:06:27.369 - Script 'ReplicatedStorage.Roact.Reconciler', Line 373 - field _reconcilePrimitiveChildren
18:06:27.370 - Script 'ReplicatedStorage.Roact.Reconciler', Line 291 - field _reconcile
18:06:27.370 - Script 'ReplicatedStorage.Roact.Component', Line 161 - method _forceUpdate
18:06:27.371 - Script 'ReplicatedStorage.Roact.Component', Line 127 - method _update
18:06:27.371 - Script 'ReplicatedStorage.Roact.Component', Line 114 - method setState
18:06:27.372 - Script 'ReplicatedStorage.RoactMaterial.Components.Checkbox', Line 35 - upvalue method
18:06:27.372 - Script 'ReplicatedStorage.Roact.SingleEventManager', Line 27
18:06:27.373 - Stack End
```

It's trivial to work around this (`condition and Roact.createElement(...) or nil`), but it would be nice if Roact would accept `boolean`s and render nothing regardless of the boolean's value. React has something similar - [React.Component.render](https://reactjs.org/docs/react-component.html#render) may return a boolean value, which results in nothing being rendered.

AmaranthineCodices:
This closes #14. It allows for the children table passed to `Roact.createElement` to contain boolean values, which will be silently ignored. This enables the use of the pattern `condition and Roact.createElement` - without this patch, if `condition` is `false`, an error is thrown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Roact Changelog
 
 ## Current master
-* No changes
+* Support booleans in reconciliation ([#14](https://github.com/Roblox/roact/issues/14))
 
 ## 1.0.0 (TODO: Date)
 * Initial release
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -223,6 +223,10 @@ function Reconciler._reifyInternal(element, parent, key, context)
 			_reifiedChildren = reifiedChildren,
 			_rbx = target,
 		}
+	elseif typeof(element) == "boolean" then
+		-- Ignore booleans of either value
+		-- See https://github.com/Roblox/roact/issues/14
+		return nil
 	end
 
 	error(("Cannot reify invalid Roact element %q"):format(tostring(element)))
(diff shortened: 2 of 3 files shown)


LPGhatguy:
Can you add a test case for the behavior? I think it might be time to create a `Reconciler.spec.lua` file for cases like this, but it can start out just describing this behavior.

LPGhatguy:
Look good! The test file looks like it's indented with four spaces instead of tabs though; once you fix that it should be set to merge!

AmaranthineCodices:
Whoops, fixed now!

PR #23, opened 2018-01-29
setState sanity checking

### LINKED ISSUE #17: Give useful error messages when calling `setState` in the wrong places
Right now, there are three places where calling `setState` is a really bad idea:
* Inside `init` (the component is not initialized and throws an unrelated error)
* Inside `render` (render needs to be pure)
* After the component is unmounted (aah!)

We should keep track of some sort of state on each component and make sure these `setState` errors are surfaced well.

AmaranthineCodices:
This fixes #17. It throws an error when setState is called in the following locations:

* the `init` function of a component
* the `render` function of a component
* any point after the component begins unmounting, **including** within `willUnmount`

The error message:

```
State cannot be set at this point: are you setting state from an init, render, or willUnmount function?
```

Tests are included, yay!

This will probably have a merge conflict with #21; I'll fix it when it happens. This required an edit to `Reconciler` to set the flag to false just before unmounting the component.

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -43,6 +43,8 @@ function Component:extend(name)
 		local self = {}
 
 		self.props = props
+		-- Used for tracking whether the component is in a position to set state.
+		self._canSetState = false
 		self._context = {}
 
 		-- Shallow copy all context values from our parent element.
@@ -63,6 +65,9 @@ function Component:extend(name)
 		if not self.state then
 			self.state = {}
 		end
+
+		-- Now that state has definitely been set, we can now allow it to be changed.
+		self._canSetState = true
 
 		return self
 	end
@@ -101,6 +106,14 @@ end
 	current state object.
 ]]
 function Component:setState(partialState)
+	-- State cannot be set in any of the following places:
+	-- * During the component's init function
+	-- * During the component's render function
+	-- * After the component has been unmounted (or is in the process of unmounting, e.g. willUnmount)
+	if not self._canSetState then
+		error("State cannot be set at this point: are you setting state from an init, render, or willUnmount function?", 0)
+	end
+
 	local newState = {}
 
 	for key, value in pairs(self.state) do
@@ -149,7 +162,9 @@ function Component:_forceUpdate(newProps, newState)
 		self.state = newState
 	end
 
+	self._canSetState = false
 	local newChildElement = self:render()
+	self._canSetState = true
 
 	if self._handle._reified ~= nil then
 		-- We returned an element before, update it.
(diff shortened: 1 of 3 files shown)


lib/Component.spec.lua
```diff
@@ -173,4 +173,56 @@ return function()
...
+			local renderElement = Core.createElement(RenderComponent)
+
+			expect(function()
+				Reconciler.reify(initElement)
+			end).to.throw()
+
+			expect(function()
+				Reconciler.reify(renderElement)
+			end).to.throw()
+		end)
```
LPGhatguy: I think we can break this one into two tests, just like how `willUnmount` is a separate test

lib/Reconciler.lua
```diff
@@ -60,6 +60,9 @@ function Reconciler.teardown(instanceHandle)
 			Reconciler.teardown(instanceHandle._reified)
 		end
 	elseif Core.isStatefulElement(element) then
+        -- Stop the component from setting state in willUnmount or anywhere thereafter.
```
LPGhatguy: Ah, there are spaces for indentation here!

lib/Component.lua
```diff
@@ -101,6 +106,14 @@ end
 	current state object.
 ]]
 function Component:setState(partialState)
+	-- State cannot be set in any of the following places:
+	-- * During the component's init function
+	-- * During the component's render function
+	-- * After the component has been unmounted (or is in the process of unmounting, e.g. willUnmount)
+	if not self._canSetState then
+		error("State cannot be set at this point: are you setting state from an init, render, or willUnmount function?", 0)
```
LPGhatguy: It might be a little clearer to say `are you calling setState from` instead of `are you setting state from`, since you can you set state from `init` as long as you don't call `setState` itself!

LPGhatguy:
Thanks!

LPGhatguy:
Oops, I should've asked you to add an entry to the changelog!

I'll go ahead and do that for both of these PRs.

LPGhatguy:
I know we've already merged this, but are there any other places where it might be inappropriate to call `setState`?

I reviewed some code today where `setState` was being called from `willUpdate`, which seems like a mistake (rendering shouldn't be re-entrant!)

AmaranthineCodices:
One place that immediately comes to mind is `shouldUpdate`. There is absolutely no reason to call `setState` from there.

`wilUpdate` and `didUpdate` are almost certainly bugs. I can think of some contrived situations where calling them wouldn't cause re-entrancy issues, but they're contrived and they're probably not worth allowing.

PR #26, opened 2018-02-01
More setState checks

AmaranthineCodices:
This builds on the discussion from #23 after it was merged. It disallows calling `setState` in *all* lifecycle hooks, not just `willUnmount`. The error message has been changed to reflect this:

```
setState cannot be used currently: are you calling setState within a lifecycle hook?
```

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -180,11 +180,11 @@ function Component:_forceUpdate(newProps, newState)
 		)
 	end
 
+	self._canSetState = true
+
 	if self.didUpdate then
 		self:didUpdate(oldProps, oldState)
 	end
-
-	self._canSetState = true
 end
 
 --[[
@@ -205,9 +205,7 @@ function Component:_reify(handle)
 	end
 
 	if self.didMount then
-		self._canSetState = false
 		self:didMount()
-		self._canSetState = true
 	end
 end
 
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -205,7 +205,9 @@ function Component:_reify(handle)
 	end
 
 	if self.didMount then
+		self._canSetState = false
 		self:didMount()
+		self._canSetState = true
 	end
 end
 
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -10,6 +10,14 @@ local Component = {}
 
 Component.__index = Component
 
+-- The error message that is thrown when setState is called in the wrong place.
+-- This is declared here to avoid really messy indentation.
+local INVALID_SETSTATE_MESSAGE = [[setState cannot be used currently, are you calling setState from any of:
+* the willUpdate or willUnmount lifecycle hooks
+* the init function
+* the render function
+* the shouldUpdate function]]
+
 --[[
 	Create a new Roact stateful component class.
 
@@ -108,7 +116,7 @@ end
 function Component:setState(partialState)
 	-- State cannot be set in any lifecycle hooks.
 	if not self._canSetState then
-		error("setState cannot be used currently: are you calling setState within a lifecycle hook?", 0)
+		error(INVALID_SETSTATE_MESSAGE, 0)
 	end
 
 	local newState = {}
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -12,7 +12,8 @@ Component.__index = Component
 
 -- The error message that is thrown when setState is called in the wrong place.
 -- This is declared here to avoid really messy indentation.
-local INVALID_SETSTATE_MESSAGE = [[setState cannot be used currently, are you calling setState from any of:
+local INVALID_SETSTATE_MESSAGE = [[
+setState cannot be used currently, are you calling setState from any of:
 * the willUpdate or willUnmount lifecycle hooks
 * the init function
 * the render function
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -106,12 +106,9 @@ end
 	current state object.
 ]]
 function Component:setState(partialState)
-	-- State cannot be set in any of the following places:
-	-- * During the component's init function
-	-- * During the component's render function
-	-- * After the component has been unmounted (or is in the process of unmounting, e.g. willUnmount)
+	-- State cannot be set in any lifecycle hooks.
 	if not self._canSetState then
-		error("setState cannot be used currently: are you calling setState from an init, render, or willUnmount function?", 0)
+		error("setState cannot be used currently: are you calling setState within a lifecycle hook?", 0)
 	end
 
 	local newState = {}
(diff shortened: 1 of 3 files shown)


lib/Component.spec.lua
```diff
@@ -207,6 +211,90 @@ return function()
...
+					a = 1
+				})
+			end
+
+			local testElement = Core.createElement(TestComponent)
+
+			expect(function()
+				local handle = Reconciler.reify(testElement)
+				-- Something of a hack, but...
+				handle._instance:shouldUpdate({}, {})
```
LPGhatguy: The other thing you can do is capture the instances `setState` function from `init`, and then call that to trigger `shouldUpdate`/`willUpdate`/`didUpdate`.

It's quite a bit more verbose, but it avoids leaking internals into tests! I don't know if it's actually better; what do you think?

lib/Component.lua
```diff
@@ -205,7 +205,9 @@ function Component:_reify(handle)
 	end
 
 	if self.didMount then
+		self._canSetState = false
 		self:didMount()
+		self._canSetState = true
```
LPGhatguy: Should we disallow `setState` in `didMount` and `didUpdate`? These might be less clear-cut.

I've definitely written code that triggers `setState` from inside those two lifecycle events, especially for code with network resources, where the flow is:
* `render` once without the data (returning nil)
* `didMount` triggers a network request on another coroutine calls `setState` to mark the component as loading
* `render` again, this time showing a spinner
* `didMount` does nothing

lib/Component.spec.lua
AmaranthineCodices: It's a lot better, because then I'm not depending on the format of the instance handles (oh no, we changed the reified handle format, our tests for setState broke!). I'll make the change.

lib/Component.lua
AmaranthineCodices: I can see use-cases for `didMount` and `didUpdate`. I can also see arguments for disallowing them, most of them centered around "avoiding footguns". I don't know that that's a very good reason to stop the sort of paradigm that you demonstrate, to be honest. Limiting the power of Roact just to avoid an error case that will *still* throw an error (stack overflow, I believe) seems like a poor decision on my part - my bad.

We should address in the documentation that you have to be careful about calling `setState` within `didMount`/`didUpdate` to avoid infinite loops - setting state in these hooks should be the *exception*, not the norm for the component's behavior.

LPGhatguy: I agree. Asynchronous rendering with an update queue would solve any of the stack overflow problems, but *does* change behavior with regards to re-rendering before your parent component finishes mounting.

lib/Component.spec.lua
```diff
@@ -207,6 +211,63 @@ return function()
...
+					a = 1
+				})
+			end
+
+			local testElement = Core.createElement(TestComponent)
+
+			expect(function()
+				Reconciler.reify(testElement)
+				shouldUpdate()
+			end).to.throw()
```
LPGhatguy: Since `shouldUpdate` is implemented entirely by the component, `_canSetState` won't be checked if we call `shouldUpdate` directly, right?

I wonder if we should call the component's `setState` instead of force updating, to hit paths like this one?

lib/Component.lua
```diff
@@ -106,12 +106,9 @@ end
...
 ]]
 function Component:setState(partialState)
-	-- State cannot be set in any of the following places:
-	-- * During the component's init function
-	-- * During the component's render function
-	-- * After the component has been unmounted (or is in the process of unmounting, e.g. willUnmount)
+	-- State cannot be set in any lifecycle hooks.
 	if not self._canSetState then
-		error("setState cannot be used currently: are you calling setState from an init, render, or willUnmount function?", 0)
+		error("setState cannot be used currently: are you calling setState within a lifecycle hook?", 0)
```
LPGhatguy: I like mentioning `render` and `init` explicitly here since they aren't lifecycle hooks.

lib/Component.spec.lua
AmaranthineCodices: This is true! It looks like this test case only passed to begin with when it re-entered `shouldUpdate`. I've fixed that now.

lib/Component.lua
AmaranthineCodices: I've changed the error message to be more descriptive:

```
setState cannot be used currently, are you calling setState from any of:
* the willUpdate or willUnmount lifecycle hooks
* the init function
* the render function
* the shouldUpdate function
```

LPGhatguy:
I've been on vacation for the past two weeks -- I'll get back into helping with this PR!

lib/Component.lua
```diff
@@ -10,6 +10,14 @@ local Component = {}
 
 Component.__index = Component
 
+-- The error message that is thrown when setState is called in the wrong place.
+-- This is declared here to avoid really messy indentation.
+local INVALID_SETSTATE_MESSAGE = [[setState cannot be used currently, are you calling setState from any of:
+* the willUpdate or willUnmount lifecycle hooks
+* the init function
+* the render function
+* the shouldUpdate function]]
```
LPGhatguy: Lua ignores the first newline after `[[` I'm fairly sure -- we should put one there to make the message more coherent!

lib/Component.lua
```diff
@@ -182,6 +188,8 @@ function Component:_forceUpdate(newProps, newState)
 		)
 	end
 
+	self._canSetState = true
```
LPGhatguy: An interesting implication here: the state of your parent can't have its state updated while you're in the middle of updating. Calling a function in your `didUpdate` callback that indirectly invokes `setState` on a parent component would thus fail.

I wonder if that's a case that we hit in real code, and if that's a bug!

lib/Component.spec.lua
```diff
@@ -184,6 +184,10 @@ return function()
 				})
 			end
 
+			function InitComponent:render()
+				return nil
+			end
```
LPGhatguy: This is an interesting artifact of a test that's only as coarse as "this high level call should throw". I wonder if there's a way to make these kinds of tests more robust in the future.

lib/Component.lua
AmaranthineCodices: This will only throw an error if this setState indirection happens synchronously. If you spawn another thread and trigger the parent `setState` call there, or do it from an event connection, there won't be an error.

Since most use-cases I can see for calling `setState` in `didUpdate` involve some sort of blocking operation (that stalls out the entire render until it's complete if it's done synchronously anyway), I don't think this is a major issue.

Hypothetically this could be worked around by deferring `didUpdate` invocations until after the entire tree has re-rendered, parents and all, but I don't think this is worth the effort.

lib/Component.lua
AmaranthineCodices: Oops, this should be fixed now.

PR #31, opened 2018-02-28
WiP: Big docs rewrite

LPGhatguy:


LPGhatguy:
This is better than what we have right now, I'm going to merge in and iterate more later.

PR #37, opened 2018-03-04
Expose access to GetPropertyChangedSignal

### LINKED ISSUE #28: Expose GetPropertyChangedSignal declaratively
Right now, you can connect to the `Changed` event in Roact easily with `Roact.Event.Changed`, but using `GetPropertyChangedSignal(name)` on a specific property is much more performant and is much cleaner.

Roact makes that painful right now -- the only way right now is to connect in `didMount` and disconnect in `willUnmount`, typical for wrappers around APIs like this.

I think we can make Roact support `GetPropertyChangedSignal` as a first-class citizen for events.

AmaranthineCodices:
Allows access to property change events using GetPropertyChangedSignal via `Roact.Change.[EventName]`. This fixes #28.

This will behave like other events; one connection per object, etc.

Note that the vast majority of this is **not unit tested**. Lemur doesn't implement GetPropertyChangedSignal x.x

Example usage:

```lua
local Roact = require(game.ReplicatedStorage.Roact)

local test = Roact.createElement("ScreenGui", {}, {
	Roact.createElement("TextBox", {
		Position = UDim2.new(0.5, 0, 0.5, 0),
		Size = UDim2.new(0, 200, 0, 40),
		[Roact.Change.Text] = function(rbx)
			print("text changed to", rbx.Text)
		end,
		Font = "SourceSans",
		TextSize = 18,
	})
})

Roact.reify(test, game.Players.LocalPlayer:WaitForChild("PlayerGui"))
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,7 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = false
+
+[*.json]
+indent_style = spaces
+indent_width = 2
--- a/rojo.json
+++ b/rojo.json
@@ -1,0 +1,14 @@
+{
+  "name": "roact",
+  "servePort": 8000,
+  "partitions": {
+    "src": {
+      "path": "lib",
+      "target": "ReplicatedStorage.Roact"
+    },
+    "testez": {
+      "path": "modules/testez/lib",
+      "target": "ReplicatedStorage.TestEZ"
+    }
+  }
+}


LPGhatguy:
Ergh. Maybe we should implement support for `GetPropertyChanged` signal before this lands. I definitely want to improve test coverage as much as possible, which will probably involve rewriting the reconciler tests.

We didn't have test coverage metrics for Lua code when the majority of the Roact tests were written! D:

AmaranthineCodices:
I would love to get support for `GetPropertyChangedSignal` in Lemur. I'm not familiar enough with Lemur's codebase and how it does events to implement it though :(

LPGhatguy:
Stuff is in Lemur!

AmaranthineCodices:
Cool! I'll try to get to this sometime this weekend.

AmaranthineCodices:
this was not intended to drag in 27 commits >.<

PR #38, opened 2018-03-04
Remove values from setState

### LINKED ISSUE #30: setState can't remove values from state
React is a JavaScript library, which means that it was designed around the existence of both `null` and `undefined`.

In React, if you want to remove a value from your state, you can use:
```js
this.setState({
    foo: null
})
```

In Lua with Roact, if you try the same thing, you'll realize it doesn't do anything:
```lua
self:setState({
    foo = nil
})
```

Lua doesn't have a way to differentiate between "nothing" and "absent" in tables! This is an API hole in Roact!

We need some way to remove a value from the state!

Two potential solutions:

* Introduce `self:removeFromState(key)`
    * This means that operations that do both a `setState` and a `removeFromState` will trigger two re-renders right now D:
* Add `Symbol` value that represents `nil`
    * There's no case where this value would end up in your state accidentally, but what would we call this symbol? `Roact.None`?

AmaranthineCodices:
This fixes #30. It adds a `Roact.None` symbol that, when used in setState, will remove the key from the state.

Usage:
```lua
-- state already contains a value named "value"
self:setState({
    value = Roact.None,
})

-- state no longer contains "value"
```

### DIFF
--- a/lib/Core.lua
+++ b/lib/Core.lua
@@ -19,7 +19,7 @@ Core.Ref = Symbol.named("Ref")
 -- Marker used to specify that a component is a Roact Portal.
 Core.Portal = Symbol.named("Portal")
 
--- Market used to specify that the value is nothing, because nil cannot be stored in tables.
+-- Marker used to specify that the value is nothing, because nil cannot be stored in tables.
 Core.None = Symbol.named("None")
 
 --[[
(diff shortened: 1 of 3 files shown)


lib/Core.lua
```diff
@@ -19,6 +19,9 @@ Core.Ref = Symbol.named("Ref")
 -- Marker used to specify that a component is a Roact Portal.
 Core.Portal = Symbol.named("Portal")
 
+-- Market used to specify that the value is nothing, because nil cannot be stored in tables.
```
LPGhatguy: Typo! s/Market/Marker

AmaranthineCodices: Oops! Fixed now.

LPGhatguy:
Looks good to me!

PR #39, opened 2018-03-05
Allow functions to be passed to setState

### LINKED ISSUE #33: Introduce setState(fn) variant
In order to implement asynchronous rendering (#18), we need to make sure there's an API in place that mirror's React's `setState` variant that accepts a function.

Dependent updates work just fine right now, which means this code that works today would not work with asynchronous rendering:

```lua
setState(self.state.value + 1)
```

Instead, we should introduce a new API and encourage users to use it today:

```lua
setState(function(props, state)
    return {
        value = state.value + 1
    }
end)
```

Once work on async rendering begins, we can start catching cases where users call `setState` multiple times in a row synchronously to try to weed out dependent state mutations.

AmaranthineCodices:
This fixes #33. It allows `setState` to accept a function, which is called with the `props` and `state` tables. This function is expected to return a partial state, which is then used to update the component's state.

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -123,7 +123,7 @@ function Component:setState(partialState)
 
 	-- If the partial state is a function, invoke it to get the actual partial state.
 	if type(partialState) == "function" then
-		partialState = partialState(self.props, self.state)
+		partialState = partialState(self.state, self.props)
 	end
 
 	local newState = {}
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -121,6 +121,11 @@ function Component:setState(partialState)
 		error(INVALID_SETSTATE_MESSAGE, 0)
 	end
 
+	-- If the partial state is a function, invoke it to get the actual partial state.
+	if type(partialState) == "function" then
+		partialState = partialState(self.props, self.state)
+	end
+
 	local newState = {}
 
 	for key, value in pairs(self.state) do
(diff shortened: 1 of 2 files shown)


lib/Component.lua
```diff
@@ -121,6 +121,11 @@ function Component:setState(partialState)
 		error(INVALID_SETSTATE_MESSAGE, 0)
 	end
 
+	-- If the partial state is a function, invoke it to get the actual partial state.
+	if type(partialState) == "function" then
+		partialState = partialState(self.props, self.state)
```
LPGhatguy: Let's flip-flop the arguments to match React. I think most cases of functional `setState` are going to depend on `state` more than `props` anyhow.

lib/Component.spec.lua
```diff
@@ -326,5 +326,43 @@ return function()
...
+			function TestComponent:render()
+				return nil
+			end
+
+			local element = Core.createElement(TestComponent)
+			local instance = Reconciler.reify(element)
+
+			expect(getStateCallback().value).to.equal(0)
+
+			setStateCallback(function(props, state)
```
LPGhatguy: We might want to assert which argument is which in this function to guard against them being the wrong tables!

lib/Component.lua
AmaranthineCodices: Makes sense; fixed now!

lib/Component.spec.lua
AmaranthineCodices: Done!

LPGhatguy:
Thanks!

PR #44, opened 2018-03-13
Add Reconciler.reconcile to the public API

### LINKED ISSUE #32: Expose reconciliation process
In the React documentation, `ReactDOM.render` is used multiple times in a single example to introduce reconciliation in a simple way. It also can simplify embedding React inside a non-React UI.

Roact requires that all state comes from within a Roact component. Calling `Roact.reify` a second time will create a new UI instead of updating the existing one.

Should Roact expose a `reconcile` API on instance handles returned from `reify`, to allow for easier embedding?

LPGhatguy:
Closes #32.

This addresses one usability issue of Roact related to updating components without having to manually expose values like `setState`.

The new API has been an internal API for awhile, and works like this:

```lua
local handle = Roact.reify(Roact.createElement("StringValue", {
    Value = "Hello",
}))

-- Previously, we were now stuck with this element and could only tear it down.
-- Now, we can modify it with Roact.reconcile:
handle = Roact.reconcile(handle, Roact.createElement("StringValue", {
    Value = "World",
}))
```

One interesting gotcha about the API is that it *takes ownership* of the handle you pass in, and has the potential to return a new one. That means that `reconcile` is free to `teardown` the entire tree that you pass it and construct a new one.

An example that gets a lot simpler is our equivalent of React's [ticking clock intro example](https://reactjs.org/docs/rendering-elements.html#updating-the-rendered-element):

```lua
local function content()
        Text = "The current time is " .. time(),
    })
end

local handle = Roact.reify(content(), ScreenGui)

while true do
    wait(1)
    handle = Roact.reconcile(handle, content())
end
```

I'll need to do a pass over the WIP documentation to recommend using this as the introductory technique for getting change. It's also useful when embedding Roact inside non-Roact components, and affords you more flexibility to own state outside the Roact tree.

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -193,7 +193,7 @@ function Component:_forceUpdate(newProps, newState)
 
 	if self._handle._reified ~= nil then
 		-- We returned an element before, update it.
-		self._handle._reified = Reconciler.reconcile(
+		self._handle._reified = Reconciler._reconcileInternal(
 			self._handle._reified,
 			newChildElement
 		)
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -193,7 +193,7 @@ function Component:_forceUpdate(newProps, newState)
 
 	if self._handle._reified ~= nil then
 		-- We returned an element before, update it.
-		self._handle._reified = Reconciler._reconcile(
+		self._handle._reified = Reconciler.reconcile(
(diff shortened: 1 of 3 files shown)


LPGhatguy:
Ugh, the coverage check is failing. I have a feeling there's going to be more refactoring sooner rather than later.

PR #46, opened 2018-03-13
Introduce global config

### LINKED ISSUE #40: New mechanism for global configuration
I've thought a lot about #25 and related issues, and I think that there might be a need for an application-level global configuration that's set once and immutable.

Particularly, I'd like to have settings like:

* Trace all Roblox mutations (#27)
* Enable traceback storage on element creation (currently `DEBUG_ENABLE`)
* Set the frame budget for async rendering
* Automatically deep-freeze props and state for debugging

The API I'd be looking for would be something like:

```lua
Roact.setGlobalConfig({
    elementTracing = true,
    deepFreezePropsAndState = true
})
```

Any calls to `setGlobalConfig` after the first would throw an error.

LPGhatguy:
Closes #40.

Introduces a single global configuration point, `elementTracing`, as well as two methods to set and get that value, `setGlobalConfig` and `getGlobalConfigValue`.


LPGhatguy:
I want to wait to merge this until #44 has landed, since it changes the public API and #44 introduces tests around what's exposed in the public API.

PR #49, opened 2018-03-14
Add symbol for element status

### LINKED ISSUE #20: More reliable detection of Roact elements
There should be a more reliable way to check if a table was returned from `createElement` than checking the `isElement` key of the resultant table. This is largely futureproofing - I want to guarantee a way to find out whether these tables actually came from `createElement`, with no false-positives (some other API sets `isElement` to true for its own reasons, for example) or possibility of breakage (I don't think Roact provides any guarantees about the structure of `createElement`'s return value).

This is probably best accomplished via a symbol - instead of checking `isElement`, I could check `[Roact.Element]` or something similar; `createElement` would set both `[Roact.Element]` and `isElement` (for backwards compatibility, if nothing else) in the table it returns.

AmaranthineCodices:
Super simple PR that fixes #20. Adds a symbol, `Roact.Element`, that is present in every table returned from `Roact.createElement`.


LPGhatguy:
Does anything check `isElement`? You might remove that flag in the same PR and update anything that used it.

AmaranthineCodices:
It's never used in Roact; I've removed it. Should be good to go!

LPGhatguy:
One more note, should we expose this symbol?

If we want to start exposing these type markers, should we have them of the form

```lua
element = {
    [Core.isElement] = true,
}
```

or of the form

```lua
element = {
    type = Core.Element,
}
```

AmaranthineCodices:
It really should be exposed - that's the point of #20. I think it's probably best to store it with the symbol as the key just to avoid colliding with anything, but really it could go either way.

LPGhatguy:
I think we get better error messages if we key with `type` consistently, since then you can tell the user what kind of thing they mistakenly passed in. It does potentially collide with other systems a bit, but I think that's okay.

AmaranthineCodices:
We can't use `type` - there's already a naming collision in the Roact codebase >.<

https://github.com/Roblox/roact/blob/340e0134b36ec9f97ddea87f2d4e2186b1eec896/lib/Core.lua#L115-L119

LPGhatguy:
Hah!

I want to change that field to be `component` anyways. `createElement` is probably the oldest function in the Roact codebase.

AmaranthineCodices:
Should I change it to that then? :p

LPGhatguy:
Yeah!

AmaranthineCodices:
Should be good to go now, though this will cause merging problems with #50.

LPGhatguy:
Conflicts are now a go! 😅

lib/Reconciler.lua
```diff
@@ -37,7 +37,7 @@ local function isPortal(element)
 		return false
 	end
 
-	return element.type == Core.Portal
+	return element.component == Core.Portal
```
LPGhatguy: Eurgh, I can't wait to rewrite the reconciler + tests so that we have better coverage for refactors like this...

PR #50, opened 2018-03-16
Remove is*Element functions

### LINKED ISSUE #45: Deprecate (and remove?) is*Element methods
I don't think methods have a place in Roact.

I don't remember why they're exposed, but they were originally just utility methods for the reconciler.

AmaranthineCodices:
Fixes #45. Moves the various element-type determining functions into `Reconciler` and removes them from the public API.

### DIFF
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -45,13 +45,46 @@ local Reconciler = {}
 Reconciler._singleEventManager = SingleEventManager.new()
 
 --[[
+	Is this element backed by a Roblox instance directly?
+]]
+local function isPrimitiveElement(element)
+	if type(element) ~= "table" then
+		return false
+	end
+
+	return type(element.type) == "string"
+end
+
+--[[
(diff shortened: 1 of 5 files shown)


AmaranthineCodices:
Coverage regression is minor - it's the untested bits (`DEBUG_ENABLE`, mostly) in `Core` that are surfacing since they aren't being diluted by the test cases for the functions.

LPGhatguy:
Those bits are going away anyways, so that's okay.

lib/Reconciler.lua
```diff
@@ -44,14 +44,47 @@ local Reconciler = {}
...
+--[[
+	Is this element defined by a component class?
+]]
+local function isStatefulElement(element)
+	if type(element) ~= "table" then
+		return false
+	end
+
+	return type(element.type) == "table"
+end
```
LPGhatguy: Ah, can we stick these up next to the `isPortal` function? It used to live in `Core` as well, but no longer does.

PR #51, opened 2018-03-17
Expose access to GetPropertyChangedSignal, take two

AmaranthineCodices:
After a hideously botched rebase attempt followed by a history rewrite happened in #37, I fixed things!

Same API as #37, but now with:

* Update lemur to the newest commit
* Fix a subtle but nasty bug that could cause property changes to almost never be connected
* Test cases! 100% test coverage in SingleEventManager.


lib/Change.lua
```diff
@@ -0,0 +1,23 @@
...
+local changeMetatable = {
+	__tostring = function(self)
+		return ("ChangeListener(%s)"):format(self.name)
+	end
+}
+
+setmetatable(Change, {
+	__index = function(self, propertyName)
+		local changeListener = {
+			type = "propertyChange",
```
LPGhatguy: We should set the type to `Change` to line up with how `Event` works.

This also lines up really nicely with the _other_ changes I want to make to typecheck arguments using the `type` value!

lib/Reconciler.lua
```diff
@@ -539,4 +541,4 @@ function Reconciler._setRbxProp(rbx, key, value, element)
 	end
 end
 
-return Reconciler
\ No newline at end of file
+return Reconciler
```
LPGhatguy: It looks like files' trailing newlines keep getting shuffled there and back, I feel like I should normalize those.

In the meantime, do you have an EditorConfig plugin enabled for your editor?

AmaranthineCodices: I do, but sometimes it doesn't work (see all the times my commits sometimes get spaces instead of tabs).

lib/Change.lua
AmaranthineCodices: Oops, I missed that part of the refactoring! Fixed now.

lib/Reconciler.lua
LPGhatguy: What editor do you use? The one for Sublime Text seems to horrifically break the automatic indentation detection. :(

AmaranthineCodices: VS Code and [this](https://github.com/editorconfig/editorconfig-vscode) extension.

AmaranthineCodices:

LPGhatguy:
Looks good to me!

PR #52, opened 2018-03-19
Trace mode

### LINKED ISSUE #27: 'Trace mode': Log all Roblox object mutations
We're finding an awesome amount of bugs with Roblox UI objects due to some projects being written using Roact. However, diagnosing the real problems with them is tricky because Roact obfuscates a lot of the actual behavior under the hood (on purpose!)

It would be useful from both a performance and debugging perspective to have a verbose logging mode for Roact to output virtual tree diffs, Roblox object mutations, and other performance notes.

AmaranthineCodices:
Fixes #27.

Adds a new configuration option, `logAllMutations`, that prints all mutations to Roblox objects and event connections. The messages look like this:

```
<TRACE> Setting BackgroundTransparency on Players.Player1.PlayerGui.ScreenGui.SomeFrame of class Frame to 0
```

And for event connections:

```
<TRACE> Connecting function to event Changed on Players.Player1.PlayerGui.ScreenGui.SomeFrame of class Frame
```

TODO:

- [x] Log mutations
- [x] Log event connections
- [x] Figure out how to test this behavior ~~(overwriting `io.write` or whatever it's called?)~~


AmaranthineCodices:
Waiting on https://github.com/LPGhatguy/lemur/pull/23 in lemur to be merged.

LPGhatguy:
Merged!

AmaranthineCodices:

AmaranthineCodices:
Err, this lemur version is not correct - it's from last November.

LPGhatguy:
That's weird. I must've really botched that upgrade somehow, should be okay now!

AmaranthineCodices:
I think you have to re-trigger the Travis build in order to see updated feedback

lib/Reconciler.lua
```diff
@@ -507,12 +509,31 @@ function Reconciler._setRbxProp(rbx, key, value, element)
...
 			error(message, 0)
+		elseif GlobalConfig.getValue("logAllMutations") then
+			local message = ("<TRACE> Setting %s on %s of class %s to %s"):format(
+				key,
+				rbx:GetFullName(),
+				rbx.ClassName,
+				tostring(value)
+			)
+
+			Reconciler._traceFunction(message)
```
LPGhatguy: I think this condition is kind of weird.

We should move it to be above the `not success` block instead of attaching an `elseif` to a block that has a negative condition. It makes sense to trace mutations even if they fail!

AmaranthineCodices: Yeah, that's a weird conditional. Fixed now!

lib/Reconciler.spec.lua
```diff
@@ -1,8 +1,87 @@
...
 
 	it("should reify booleans as nil", function()
 		local booleanReified = Reconciler.reify(false)
 		expect(booleanReified).to.never.be.ok()
 	end)
+
+	describe("tracing", function()
+		GlobalConfig.set({
+			logAllMutations = true
+		})
```
LPGhatguy: This is super flaky. `GlobalConfig` is global for a reason, and if any other tests try to set it, one of them will randomly fail.

In each `it` block, you should set + reset the config value, which also gives you an opportunity to test the flag unset case, too!

AmaranthineCodices: Oops, yeah. Fixed now!

LPGhatguy:
Hack week is over!

I'll revisit this PR this week; I haven't forgotten.

LPGhatguy:
I pulled this PR down to try it, and instance naming doesn't work during the reify stage!

None of the instances are parented until they're fully constructed, including any of their children, meaning that whenever there's a new instance created, we don't know where it is or what it is.

AmaranthineCodices:
How should that be handled? Deferring the trace statements until after parenting?

LPGhatguy:
We might be able to reconstruct what the full name of the instance _will_ be before we parent it.

AmaranthineCodices:
We...can, but it requires passing a *lot* of information through side channels, because of how we construct children, unless I'm really overthinking things?

LPGhatguy:
We can make the component instance keep a reference to its parent instance, which should be safe because instances are never relocated. Each component along the tree can potentially have a `_rbx` property, and you can crawl upwards and append each element's name to a list.

Keeping a persistent map of instances is kind of flaky because it is side-channel data, and it's global!

AmaranthineCodices:
These tests fail on lemur but run in Studio - I'm not entirely sure what's going on here. Somehow `GlobalConfig` is not being reset?

LPGhatguy:
I'll take a look. It's possible the tests are running in a different order and that's messing things up.

lib/Reconciler.lua
```diff
@@ -149,39 +151,52 @@ end
...
+			rbx.Name = key
+		end
+
+		-- Only set these properties if tracing is enabled, since that's what
+		-- these are used for.
+		if GlobalConfig.getValue("logAllMutations") then
+			element._rbx = rbx
+			element._parent = parentElement
+			element._rbxParent = parent
+		end
```
LPGhatguy: This doesn't actually work -- you can reuse elements to create multiple trees and breaks the immutability of elements in a pretty bad way.

We should be modifying the component *instances* or *instance handles*. They already have these kinds of values and are stateful and mutable.

AmaranthineCodices:
Okay, test cases are erroring somewhere; I didn't catch this originally. The test case sets `logAllMutations` to true, then errors, so `GlobalConfig` is never reset. Tracking down the failing case now >.<

LPGhatguy:
I promise I haven't forgotten about this, I've just been swamped with other stuff v.v

AmaranthineCodices:
Any word on this?

lib/Reconciler.lua
```diff
@@ -467,6 +483,52 @@ local function set(rbx, key, value)
...
+					fullName = level._parent:GetFullName() .. "." .. fullName
+				end
+
+				break
+			end
+		end
+
+		return fullName
+	end
+end
```
LPGhatguy: My gut says that this function can probably be much simpler.

If we add `_parentHandle` to every node (which makes sense to me), then we can have something like:

```lua
local function computeFullName(handle)
	assert(handle._rbx, "Can only computeFullName of a primitive instance handle!")

	local fullName = handle._key

	while handle._parentHandle do
		handle = handle._parentHandle

		if handle._rbx then
			fullName = handle._key .. "." .. fullName
		end
	end

	if handle._parent then
		fullName = handle._parent:GetFullName() .. fullName
	end

	return fullName
end
```

lib/Reconciler.lua
```diff
@@ -479,6 +487,30 @@ local function set(rbx, key, value)
 	rbx[key] = value
 end
 
+--[[
+	Computes a full name, even if the element is not parented to the DataModel yet.
+]]
+local function computeFullName(handle)
+	-- Assert that this is a primitve handle.
```
LPGhatguy: I don't think this comment adds very much information!

LPGhatguy:
Let's revisit this i nthe context of the new reconciler later!

PR #57, opened 2018-04-04
Add getDerivedStateFromProps

### LINKED ISSUE #53: Add getDerivedStateFromProps
React provides a lifecycle hook named [componentWillReceiveProps](https://reactjs.org/docs/react-component.html#componentwillreceiveprops). It is invoked with the next property table, which may or may not be equal to the current one. Crucially, it *allows `setState` to be called* within it, allowing components to respond to props with a change in state.

This would be very beneficial to have in Roact. Potential use-cases include:

* Changing a button's `hover` state if its position changes
* Resizing a component (implementing your own `SizeFromContents`, for example) in response to property changes
* Fixing https://github.com/Roblox/roact-rodux/issues/2

AmaranthineCodices:
Fixes #53.

Adds a static class function, `getDerivedStateFromProps`, that is invoked when a component instance is created and when properties change. This function receives two arguments: `newProps`, the new properties table, and `currentState`, the component's current state (note that in cases where rendering causes both properties and state to change, `currentState` will equal the new state - I'm not sure if this happens in practice).

`getDerivedStateFromProps` may return either a table or `nil`; if it returns a table, the table is merged into the state without triggering another re-render. This is called before any lifecycle hooks invoke, including `willUpdate`.

### DIFF
--- a/lib/Component.spec.lua
+++ b/lib/Component.spec.lua
@@ -177,74 +177,43 @@ return function()
 		Reconciler.teardown(instance)
 	end)
 
-	describe("getDerivedStateFromProps", function()
-		it("should be called when a component is initialized", function()
-			local TestComponent = Component:extend("TestComponent")
-			local getStateCallback
+	it("should call getDerivedStateFromProps appropriately", function()
+		local TestComponent = Component:extend("TestComponent")
+		local getStateCallback
 
-			function TestComponent.getDerivedStateFromProps(newProps, oldState)
-				return {
-					visible = newProps.visible
-				}
+		function TestComponent.getDerivedStateFromProps(newProps, oldState)
+			return {
+				visible = newProps.visible
+			}
+		end
+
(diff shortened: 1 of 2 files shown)


lib/Component.spec.lua
```diff
@@ -177,6 +177,76 @@ return function()
...
+				getStateCallback = function()
+					return self.state
+				end
+			end
+
+			function TestComponent:render() end
+
+			local handle = Reconciler.reify(Core.createElement(TestComponent, {
+				visible = true
+			}))
```
LPGhatguy: We can merge these two tests just by adding the two lines of assertion here I think.

AmaranthineCodices: Fixed now!

LPGhatguy:
I'll handle the changelog and documentation -- I want to hold off on the latter until I actually get a chance to try porting Roact-Rodux to use this!

PR #62, opened 2018-04-10
WIP: Add instrumentation to reconciler to track wasted renders

ZoteTheMighty:
Wasted renders may be contributing significantly to performance issues in Roact apps. This is a pretty basic check to get high level stats on how much extra work the UI is doing.  Hopefully this can be expanded to provide more specific information as well.

### DIFF
--- a/lib/DebugTrackWastedRenders.lua
+++ b/lib/DebugTrackWastedRenders.lua
@@ -1,0 +1,115 @@
+local DebugTrackWastedRenders = {}
+
+-- Tracks total number of renders by type of component
+local totalRenders = 0
+local wastedRendersByType = {}
+local validRendersByType = {}
+
+-- Allows tracking of frames without renders, so we can only report when needed
+local previousTotal = 0
+local framesWithoutUpdate = 0
+-- TODO: Make this configurable somehow without overcomplicating GlobalConfig
+local renderFramesTillSettled = 10
+
+local HeartbeatConnection = nil
+
+--[[
+	Determine if an old and new elements are equal. Used to detect wasted renders
+	when the trackWastedRenders flag is set
+]]
+local function deepEquals(old, new)
+	if old == new then
+		return true
+	end
+
+	if (type(old) == "table" and type(new) == "table") then
+		for key, oldVal in pairs(old) do
+			if not deepEquals(new[key], oldVal) then
+				return false
+			end
+		end
+
+		for key, newVal in pairs(new) do
+			if not deepEquals(newVal, old[key]) then
+				return false
+			end
+		end
+	else
+		return false
+	end
+
+	return true
+end
+
+
+--[[
+	Print accumulated wasted render stats once the number of renders stops increasing
+	for the amount of frames specified in the config
+]]
+local function printStatsWhenSettled()
+	if previousTotal == totalRenders then
+		framesWithoutUpdate = framesWithoutUpdate + 1
+
+		-- If it's been enough frames without new Roact Renders, report results
+		if framesWithoutUpdate == renderFramesTillSettled then
+			-- TODO: More precise/accurate messaging, better spacing
+			-- Print stats on wasted renders
+			print("UI settled. Render stats:")
+			for type, _ in pairs(wastedRendersByType) do
+				local wasted = wastedRendersByType[type]
+				local valid = validRendersByType[type]
+				if wasted > 0 or valid > 0 then
+					print((
+						"	%s: %d wasted, %d valid; Amount valid = %d%%"
+					):format(type, wasted, valid, valid / (wasted + valid) * 100))
+				end
+			end
+			-- Clear the wasted render data
+			totalRenders = 0
+
+			for k, _ in pairs(wastedRendersByType) do
+				wastedRendersByType[k] = 0
+				validRendersByType[k] = 0
+			end
+		end
+	else
+		-- Reset the counter as long as renders are still happening
+		framesWithoutUpdate = 0
+	end
+	previousTotal = totalRenders
+end
+
+--[[
+	Start start reporting wasted renders whenever Roact settles
+]]
+function DebugTrackWastedRenders.startReporting()
+	HeartbeatConnection = game:GetService("RunService").Heartbeat:connect(function()
+		printStatsWhenSettled()
+	end)
+end
+
+--[[
+	Stop reporting. Note that this doesn't affect whether or not wasted renders are tracked
+]]
+-- TODO: Find out if/where this should be called, possibly better align tracking and reporting
+function DebugTrackWastedRenders.stopReporting()
+	if HeartbeatConnection then
+		HeartbeatConnection:Disconnect()
+		HeartbeatConnection = nil
+	end
+end
+
+--[[
+	When Roact renders an element, it will report it via this method. If the
+	new element is equivalent to the old one, we consider it a wasted render
+]]
+function DebugTrackWastedRenders.recordRender(type, newElement, oldElement)
+	totalRenders = totalRenders + 1
+	if deepEquals(oldElement, newElement) then
+		wastedRendersByType[type] = (wastedRendersByType[type] or 0) + 1
+	else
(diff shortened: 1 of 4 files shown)


lib/Reconciler.lua
```diff
@@ -306,6 +312,22 @@ function Reconciler._reconcileInternal(instanceHandle, newElement)
...
+	-- Check for wasted renders if the config is set
+	if GlobalConfig.getValue("trackWastedRenders") then
+		local type
+		if isPrimitiveElement(newElement) then
+			type = "Primitive"
+		elseif isFunctionalElement(newElement) then
+			type = "Functional"
+		elseif isStatefulElement(newElement) then
+			type = "Stateful"
+		else
```
AmaranthineCodices: This seems like it'd break if we add another kind of element in the future. Could we change this `else` to `elseif isPortal(newElement)`, and warn (error?) if we couldn't determine the type?

lib/DebugTrackWastedRenders.lua
```diff
@@ -0,0 +1,115 @@
...
+--[[
+	Determine if an old and new elements are equal. Used to detect wasted renders
+	when the trackWastedRenders flag is set
+]]
+local function deepEquals(old, new)
+	if old == new then
+		return true
+	end
+
+	if (type(old) == "table" and type(new) == "table") then
```
AmaranthineCodices: nitpick: the conditional shouldn't be wrapped in parentheses

lib/DebugTrackWastedRenders.lua
```diff
@@ -0,0 +1,115 @@
...
+local wastedRendersByType = {}
+local validRendersByType = {}
+
+-- Allows tracking of frames without renders, so we can only report when needed
+local previousTotal = 0
+local framesWithoutUpdate = 0
+-- TODO: Make this configurable somehow without overcomplicating GlobalConfig
+local renderFramesTillSettled = 10
+
+local HeartbeatConnection = nil
```
AmaranthineCodices: nitpick: variable names should be `camelCase`

lib/DebugTrackWastedRenders.lua
```diff
@@ -0,0 +1,115 @@
...
+]]
+local function printStatsWhenSettled()
+	if previousTotal == totalRenders then
+		framesWithoutUpdate = framesWithoutUpdate + 1
+
+		-- If it's been enough frames without new Roact Renders, report results
+		if framesWithoutUpdate == renderFramesTillSettled then
+			-- TODO: More precise/accurate messaging, better spacing
+			-- Print stats on wasted renders
+			print("UI settled. Render stats:")
```
AmaranthineCodices: Can we change `UI` to `Roact rendering` or something similar? Sometimes Roact gets used for non-UI applications, so this message could be confusing!

lib/DebugTrackWastedRenders.spec.lua
```diff
@@ -0,0 +1,8 @@
+return function()
+
+	describe("wastedRenderTracking", function()
+		it("should print each time a set of renders completes", function ()
+			-- TODO: Figure out how to test this given that it connects to heartbeat
```
AmaranthineCodices: This will probably require knowledge of the internal state of the tracker; I think it's okay to expose that, since we're not exporting the module in `init.lua`! With that said, testing this right now is not possible, since lemur doesn't have an event loop for events to use.

lib/DebugTrackWastedRenders.spec.lua
```diff
@@ -0,0 +1,8 @@
+return function()
+
+	describe("wastedRenderTracking", function()
+		it("should print each time a set of renders completes", function ()
```
AmaranthineCodices: nitpick: space between `function` and `()`

lib/DebugTrackWastedRenders.lua
AmaranthineCodices: This message gets printed even if there are no render stats to report!

lib/DebugTrackWastedRenders.lua
```diff
@@ -0,0 +1,115 @@
...
+			for type, _ in pairs(wastedRendersByType) do
+				local wasted = wastedRendersByType[type]
```
AmaranthineCodices: If either of these are `nil`, an error will be thrown.

lib/DebugTrackWastedRenders.lua
```diff
@@ -0,0 +1,115 @@
+local DebugTrackWastedRenders = {}
+
+-- Tracks total number of renders by type of component
+local totalRenders = 0
+local wastedRendersByType = {}
```
AmaranthineCodices: To make this less error-prone, let's make this have a default of 0 with metatables:

```lua
local wastedRendersByType = setmetatable({}, {
    __index = function(self, key)
        self[key] = 0
        return 0
    end,
})
```

and the same for `validRendersByType`.

AmaranthineCodices:
There's also some kind of weird behavior here. This code will report that 50% of all primitive renders are wasted:

```lua
local Roact = require(game.ReplicatedStorage.Roact)
Roact.setGlobalConfig({
	trackWastedRenders = true
})

local TestComponent = Roact.Component:extend("Test")

function TestComponent:init()
	self.state = {
		pos = 1
	}
end

function TestComponent:render()
		Position = UDim2.new(0, self.state.pos, 0, 0),
	}, {
		Roact.createElement("Frame", {
			Position = UDim2.new(0, 1, 0, 0)
		})
	})
end

function TestComponent:didMount()
	spawn(function()
		while true do
			wait(1)
			self:setState({
				pos = math.random(-100, 100),
			})
		end
	end)
end

Roact.reify(Roact.createElement(TestComponent), workspace)
```

This isn't the best behavior - while the parent is changing, the child is exactly the same, and Roact ends up only setting a property, since that's the only thing that changed. This code implies that the child's render is wasted, but it's not - it's exactly where it should be.

lib/DebugTrackWastedRenders.lua
LPGhatguy: The other thing we can do is just default initialize each component type manually, which would make it so we can still iterate over the table.

ZoteTheMighty:
Thanks for the feedback!  I'm going to start with finding a way to adjust this to account for what @AmaranthineCodices mentioned in regards to primitive elements; it may be better to disregard them altogether, since it's stateful/functional components we're really concerned with.

It may also be best to focus on building in tracking capabilities without necessarily attempting to log them in any intelligent way, but later integrate them into a more full-featured dev tool.

AmaranthineCodices:
Worth noting: that's not exclusive to primitive elements. This has the same incorrect behavior:

```lua
local Roact = require(game.ReplicatedStorage.Roact)
Roact.setGlobalConfig({
	trackWastedRenders = true
})

local function Dummy(props)
		Position = UDim2.new(0, 1, 0, 0)
	})
end

local TestComponent = Roact.Component:extend("Test")

function TestComponent:init()
	self.state = {
		pos = 1
	}
end

function TestComponent:render()
		Position = UDim2.new(0, self.state.pos, 0, 0),
	}, {
		Roact.createElement(Dummy)
	})
end

function TestComponent:didMount()
	spawn(function()
		while true do
			wait(1)
			self:setState({
				pos = math.random(-100, 100),
			})
		end
	end)
end

Roact.reify(Roact.createElement(TestComponent), workspace)
```

With the output being
```
UI settled. Render stats:
	Functional: 1 wasted, 0 valid; Amount valid = 0%
	Primitive: 1 wasted, 1 valid; Amount valid = 50%
```

I had some more thoughts on this.

Firstly, I'm not sure that outputting render stats when renders cease is the best solution. We have all the information to know *immediately* whether a render is wasted, so we can output it immediately with a stack trace (if `elementTracing` is enabled). This will help with locating unnecessary renders. It will also allow error checking when Roact is re-rendering something faster than the current frame limit - right now this prints nothing when that happens!

Secondly, on the subject of "what is a wasted render"...it's complicated. I'm not entirely sure how to differentiate between "supposed to have the same properties every time" and "wasted render/bug" without extra hinting.

PR #64, opened 2018-04-11
Allow canceling render

### LINKED ISSUE #63: setState should allow functional arguments to return nil
This is a feature that I really wanted when working on Roact-Rodux. I overlooked it in the initial pull request!

See: https://x-team.com/blog/react-render-setstate/

AmaranthineCodices:
Tiny patch that fixes #63. Unit test added to confirm it works!


LPGhatguy:
Yay!

PR #65, opened 2018-04-12
Component benchmarks

ZoteTheMighty:
Retrying my approach to tracking performance. This will gather stats about `render` calls and `shouldUpdate` calls, and should be able to provide info that would help indicate when to make a `Component` a `PureComponent` or vice versa, when to write a custom `shouldUpdate`, or when to try to cut down on the complexity of `render` methods.

I didn't find a particularly clean way to inject the instrumentation into the `Component` code without muddling it a bit, so I'd love suggestions for integrating it more gracefully (I left TODOs to this effect).

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -6,9 +6,15 @@
 
 local Reconciler = require(script.Parent.Reconciler)
 local Core = require(script.Parent.Core)
+local GlobalConfig = require(script.Parent.GlobalConfig)
+local Instrumentation = require(script.Parent.Instrumentation)
+
 local invalidSetStateMessages = require(script.Parent.invalidSetStateMessages)
 
 local Component = {}
+
+-- Locally cache tick so we can minimize impact of calling it for instrumentation
+local l_tick = tick
 
 Component.__index = Component
 
@@ -180,7 +186,21 @@ end
 ]]
 function Component:_update(newProps, newState)
 	self._setStateBlockedReason = "shouldUpdate"
-	local doUpdate = self:shouldUpdate(newProps or self.props, newState or self.state)
+
+	-- TODO: Encapsulate the instrumentation more gracefully
+	local doUpdate
+	if GlobalConfig.getValue("shouldUpdateInstrumentation") then
+		-- Start timing
+		local time = l_tick()
+		doUpdate = self:shouldUpdate(newProps or self.props, newState or self.state)
+		-- Finish timing
+		time = l_tick() - time
+		-- Log result
+		Instrumentation.logShouldUpdate(self._handle, doUpdate, time)
+	else
+		doUpdate = self:shouldUpdate(newProps or self.props, newState or self.state)
+	end
+
 	self._setStateBlockedReason = nil
 
 	if doUpdate then
@@ -228,7 +248,19 @@ function Component:_forceUpdate(newProps, newState)
 	end
 
 	self._setStateBlockedReason = "render"
-	local newChildElement = self:render()
+	-- TODO: Encapsulate the instrumentation more gracefully
+	local newChildElement
+	if GlobalConfig.getValue("renderInstrumentation") then
+		-- Start timing
+		local time = l_tick()
+		newChildElement = self:render()
+		-- End timing
+		time = l_tick() - time
+		-- Log result
+		Instrumentation.logRenderTime(self._handle, time)
+	else
+		newChildElement = self:render()
+	end
 	self._setStateBlockedReason = nil
 
 	self._setStateBlockedReason = "reconcile"
@@ -262,7 +294,21 @@ function Component:_reify(handle)
 	self._handle = handle
 
 	self._setStateBlockedReason = "render"
-	local virtualTree = self:render()
+
+	-- TODO: Encapsulate the instrumentation more gracefully
+	local virtualTree
+	if GlobalConfig.getValue("renderInstrumentation") then
+		-- Start timing
+		local time = l_tick()
+		virtualTree = self:render()
+		-- End timing
+		time = l_tick() - time
+		-- Log result
+		Instrumentation.logRenderTime(self._handle, time)
+	else
+		virtualTree = self:render()
+	end
+
 	self._setStateBlockedReason = nil
 
 	if virtualTree then
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -14,7 +14,7 @@ local invalidSetStateMessages = require(script.Parent.invalidSetStateMessages)
 local Component = {}
 
 -- Locally cache tick so we can minimize impact of calling it for instrumentation
-local l_tick = tick
+local tick = tick
 
 Component.__index = Component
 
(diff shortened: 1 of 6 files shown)


lib/Component.lua
```diff
@@ -6,9 +6,14 @@
 
 local Reconciler = require(script.Parent.Reconciler)
 local Core = require(script.Parent.Core)
+local GlobalConfig = require(script.Parent.GlobalConfig)
+local Instrumentation = require(script.Parent.Instrumentation)
 
 local Component = {}
 
+-- Locally cache tick so we can minimize impact of calling it for instrumentation
+local l_tick = tick
```
LPGhatguy: We can just call our local version `tick`, since Lua lets you do that.

lib/Instrumentation.lua
```diff
@@ -0,0 +1,183 @@
...
+end
+
+function Instrumentation.printStats(sortBy)
+	sortBy = sortBy or "component"
+
+	local trackUpdates = GlobalConfig.getValue("shouldUpdateInstrumentation")
+	local trackRenders = GlobalConfig.getValue("renderInstrumentation")
+
+	if not trackUpdates and not trackRenders then
+		print("No stats are being tracked!  Enable with Roact.setConfig({...})")
```
AmaranthineCodices: `Roact.setGlobalConfig`, not `Roact.setConfig`!

lib/Instrumentation.lua
```diff
@@ -0,0 +1,183 @@
...
+				"%-12d%-12.3f%-12.3f"
+			):format(
+				stat.renderCount,
+				stat.avgRenderTime,
+				stat.renderPct
+			)
+		end
+		print(statsRow)
+	end
+	if trackUpdates then
```
AmaranthineCodices: Are these stats terribly useful, given that they aren't attributed to specific components?

lib/init.lua
```diff
@@ -39,6 +40,7 @@ apply(Roact, {
 	PureComponent = PureComponent,
 	Event = Event,
 	Change = Change,
+	Instrumentation = Instrumentation,
```
AmaranthineCodices: We can probably only expose `printStats` here instead of exposing the entire module.

lib/Instrumentation.lua
```diff
@@ -0,0 +1,183 @@
...
+		-- Percent of total shouldUpdate time by component
+
+local function valuesByStat(t, f)
+	local a = {}
+	for _, obj in pairs(t) do table.insert(a, obj) end
+	table.sort(a, f)
+	local i = 0      -- iterator variable
+	local iter = function ()   -- iterator function
+		i = i + 1
+		if a[i] == nil then
```
AmaranthineCodices: This entire if block can be simplified to `return a[i]`.

lib/Instrumentation.lua
```diff
@@ -0,0 +1,183 @@
...
+		-- Average shouldUpdate time by component
+		-- Percent of time shouldUpdate returns true
+		-- Percent of total shouldUpdate time by component
+
+local function valuesByStat(t, f)
+	local a = {}
+	for _, obj in pairs(t) do table.insert(a, obj) end
+	table.sort(a, f)
+	local i = 0      -- iterator variable
+	local iter = function ()   -- iterator function
```
AmaranthineCodices: nitpick: no spaces between `function` and parentheses

lib/Instrumentation.lua
```diff
@@ -0,0 +1,183 @@
...
+	if handle and handle._element and handle._element.component then
+		name = tostring(handle._element.component)
+	else
+		warn("Component name not valid for " .. tostring(handle._key))
+		return nil
+	end
+	local entry = ComponentStats[name]
+	if not entry then
+		-- Use shortened name for convenience; 
+		local shortName = name:gsub("Connection","Conn")
```
AmaranthineCodices: Let's not shorten names like this - finding poorly-performing components could be tricky if the name gets squashed. If we absolutely must shorten names, we could try trimming the string after it passes a specific number of characters (maybe trim from the end, so `ReallyLongName` gets trimmed to something like `...ongName`? Not sure about the exact semantics).

AmaranthineCodices:
No, it's okay! I started reviewing before you pushed the second commit, and didn't see it before I submitted the review. Don't worry about squashing commits; @LPGhatguy will do that while merging. This looks good as of now, barring lpg's note about `tick`.

lib/init.lua
```diff
@@ -46,4 +47,9 @@ apply(Roact, {
 	getGlobalConfigValue = GlobalConfig.getValue,
 })
 
+apply(Roact, {
+	getCollectedStats = Instrumentation.getCollectedStats,
+	clearCollectedStats = Instrumentation.clearCollectedStats,
+})
```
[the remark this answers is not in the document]
ZoteTheMighty: That's fair. I've addressed it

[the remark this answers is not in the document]
AmaranthineCodices: Would it be okay to expose them as a table, something like `Roact.Instrumentation.getCollectedStats`?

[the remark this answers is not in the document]
ZoteTheMighty: Heh, sounds my original implementation.  @LPGhatguy It would be good to expose them in some way so that there's some utility to these changes, especially for the sake of the other tooling I'm working on.  Let me know what you think.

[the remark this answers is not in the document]
LPGhatguy: Putting them at `Roact.Instrumentation.getCollectedStats` would be fine.

The original version exposed them as `Roact.getCollectedStats`, that was the problematic variant!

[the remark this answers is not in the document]
ZoteTheMighty: The original version (a few commits ago) actually had the whole `Instrumentation` object exposed, but it makes more sense to expose it indirectly.

ZoteTheMighty:
@LPGhatguy Any other changes you'd like to see before merging this?

lib/Config.lua
```diff
@@ -16,6 +16,10 @@
 local defaultConfig = {
 	-- Enables storage of `debug.traceback()` values on elements for debugging.
 	["elementTracing"] = false,
+	-- Enables instrumentation of shouldUpdate
+	["shouldUpdateInstrumentation"] = false,
+	-- Enables instrumentation of rendering
+	["renderInstrumentation"] = false,
```
LPGhatguy: Do you think it's valuable to toggle instrumentation like this individually?

When working with a profiling debugger, I imagine we would either want all of the information logged, or none of it!

ZoteTheMighty: I agree wholeheartedly, especially now that I'm faced with the idea of having my plugin behave accordingly if something is disabled. I'll make that change.

LPGhatguy:
Okay, I'm moving forward with the merge now because this PR only minor changes to the actual reconciler code, and doesn't affect the public API (since it's marked UNSTABLE). That should give us flexibility to modify and improve the stats that we report going forward without worrying about developers depending on this API.

PR #67, opened 2018-04-12
Introduce much more detailed setState error messages

### LINKED ISSUE #66: Improve setState error messages
The current error message ended up being super confusing when it's used for _every_ case where `setState` is locked.

I'm going to refactor the code to introduce special error messages for each case where setState is locked.

LPGhatguy:
Fixes #66.

This PR replaces the single "you can't use setState here" message with a bunch of new messages, each tailored to the exact case.

They give much more useful information, and where possible, where to put your `setState` call instead.

It also catches an extra case where the render and reconciliation steps _were not_ being guarded, specifically during reify.

### DIFF
--- a/lib/invalidSetStateMessages.lua
+++ b/lib/invalidSetStateMessages.lua
@@ -28,7 +28,13 @@ Check the definition of shouldUpdate in the component %q.]]
 
 invalidSetStateMessages["init"] = [[
 setState cannot be used in the init method.
-Consider using the didMount method instead.
+During init, the component hasn't initialized yet, and isn't ready to render.
+
+Instead, set the `state` value directly:
+
+	self.state = {
+		value = "foo"
+	}
 
 Check the definition of init in the component %q.]]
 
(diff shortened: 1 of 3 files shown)


lib/invalidSetStateMessages.lua
```diff
@@ -0,0 +1,59 @@
...
+
+invalidSetStateMessages["reconcile"] = [[
+setState cannot be called while a component is being reified or reconciled.
+This is the step where Roact constructs Roblox instances, and starting another
+render here would introduce bugs.
+
+Check the component %q to see if:
+	* setState could be called by a child Ref
+	* setState could be called by a child Changed event
+	* setState could be called by a child's render method]]
```
ZoteTheMighty: Nitpick, but might be clearer to say "setState is being called by...".  Otherwise, it kind of sounds like a suggestion rather than a possible diagnosis.

LPGhatguy: I'll probably also fold that part into the section before the list, like:

Check the component %q to see if setState is being called by:
* a child Ref
* a child Changed event
* a child's render method

lib/invalidSetStateMessages.lua
```diff
@@ -0,0 +1,59 @@
...
+Check the definition of willUnmount in the component %q.]]
+
+invalidSetStateMessages["shouldUpdate"] = [[
+setState cannot be used in the shouldUpdate lifecycle method.
+shouldUpdate must be a pure function that only depends on props and state.
+
+Check the definition of shouldUpdate in the component %q.]]
+
+invalidSetStateMessages["init"] = [[
+setState cannot be used in the init method.
```
AmaranthineCodices: It might be helpful to mention that you can set `state` directly in `init`.

LPGhatguy:
All of the messages are updated.

Any other feedback?

ZoteTheMighty:
Nah, LGTM!

AmaranthineCodices:
LGTM as well

PR #72, opened 2018-04-23
Make test for public API more strict

LPGhatguy:
There were a couple APIs that we were exposing that didn't have tests to validate them.

This changes the `init.spec.lua` API to require that everything we expose is explicit and intentionally exposed.


ZoteTheMighty:
WRT unstable parts of the API: would the idea be to have a something like an `UNSTABLE` child of the main Roact object?  Then just set to `UNSTABLE = true` in the public api object?

LPGhatguy:
That would work, yeah.

PR #73, opened 2018-04-24
Introduce benchmarking system

LPGhatguy:
This needs the latest Rojo master, as well as a change to Lemur to work right now.

Once all those things are merged, and you've installed LuaSocket, you can do this:

```
$ lua bench.lua
Starting        2       benchmarks...
Benchmark hello.bench:
        (100000 iterations) took 1.424000 ms (14.240000 ns/iteration)
Benchmark reconcile.bench:
        (100000 iterations) took 0.465000 ms (4.650000 ns/iteration)
Benchmarks complete!
```

The benchmark 'server' script works inside Roblox provided that your paths are set up correctly. This requires a fix in Rojo that allows partitions to target files directly. I didn't feel like working around that for now.

### DIFF
--- a/bench.lua
+++ b/bench.lua
@@ -45,22 +45,22 @@ end
 local habitat = lemur.Habitat.new()
 
 -- We'll put all of our library code and dependencies here
-local Root = habitat.game:GetService("ReplicatedStorage")
+local root = habitat.game:GetService("ReplicatedStorage")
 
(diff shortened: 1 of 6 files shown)


LPGhatguy:
This code should work as-is now. It also introduces a fix that should let us upgrade to the latest version of Lemur!

PR #74, opened 2018-04-27
WIP docs improvements, performance guide

ZoteTheMighty:
Some work-in-progress documentation changes. There's a lot to do still, but some of these things may be worth adding in even incrementally.

I'd also like to get eyes on this before I write too much of it, so I can get feedback.


docs/advanced/performance.md
```diff
@@ -0,0 +1,67 @@
+# Performance Guide
+Roact operates on the principle that it's much easier to build UI declaratively. When something changes, the UI is rebuilt virtually and then the minimal necessary changes are made to the actual UI elements.
+
+For simple projects, performance is unlikely to be an issue. But Roact is built in Lua, and Lua can be slow. There are a number of techniques that you can do to avoid performance strain with your Roact apps.
+
+## Reduce Reconciliation
+
+### `shouldUpdate` Lifecycle Method
+When a Roact Component's state or props change, it will call the Component's `shouldUpdate` method to determine whether or not to re-render it. The default implementation will always returns true.
```
SlartibartfastFjords: Typo. Should be "return"

docs/advanced/performance.md
```diff
@@ -0,0 +1,67 @@
...
+```lua
+function Component:shouldUpdate(newProps, newState)
+	return true
+end
+```
+
+If you have a more complex component that only needs to re-render in certain situations, you can implement `shouldUpdate` and return `false` in any case where an update is not required.
```
SlartibartfastFjords: Might want to mention PureComponent before providing this alternative.

docs/advanced/performance.md
```diff
@@ -0,0 +1,67 @@
...
+function Item:render()
+	local icon = self.props.icon
+
+	-- Create an ImageLabel with the item's icon
+	return Roact.createElement("ImageLabel", {
+		Image = icon,
+	})
+end
+
+local Inventory = Roact.Component:extend("FriendList")
```
SlartibartfastFjords: Name mismatch

docs/advanced/performance.md
```diff
@@ -0,0 +1,67 @@
...
+
+Lets change `Item` to a `PureComponent`:
+```lua
+local Item = Roact.PureComponent:extend("Item")
+```
+Now, if we add a new item to the `Inventory` or change something about an existing item, we'll only re-render the `Inventory` itself and the modified `Item`!
+
+### Immutable Props
+
+Immutable props are critical to properly using `PureComponent`. They ensure that any prop that is a list or a table will always change when its contents change.
```
SlartibartfastFjords: "They ensure that any prop that is a list or a table will always change when its contents change" is a little vague. Maybe mention that immutability guarantees that the reference to prop changes whenever the prop's contents are changed?

docs/advanced/performance.md
LPGhatguy: We should also warn people that implementing `shouldUpdate` is _dangerous_!

docs/advanced/performance.md
ZoteTheMighty: I like that better.  I think it's still worth mentioning that changing lists and tables are the primary concrete examples. I'd like to add a whole section on immutability, with some examples of how you can break PureComponent by not using it.

ZoteTheMighty:
The last changes I made split up the perf docs a little bit, but they may not diff nicely with the previous commit. Sorry about that :/

docs/advanced/refs.md
```diff
@@ -1,7 +1,59 @@
...
+
+## Refs in Action
+
+To create a ref, pass a function prop with the key `Roact.Ref` when creating a primitive element.
+
+For example, suppose we wanted to create a search bar that captured cursor focus when any part of it was clicked. We might use a component like this:
+
+```lua
+--[[ A search bar with an icon and a text box that captures focus for 
+ its TextBox when its icon is clicked ]]
```
LPGhatguy: As a minor style note (and something we should include in the style guide!), multi-line comments should generally start/end on their own line:

```lua
--[[
	A search bar with an icon and a text box that captures focus for its TextBox
	when its icon is clicked.
]]
```

Sublime's auto-wrap will fight you on this unless you explicitly select just the text before you press `Alt+Q`, which is kind of a shame.

docs/advanced/refs.md
```diff
@@ -1,7 +1,59 @@
...
+local SearchBar = Roact.Component:extend("SearchBar")
+
+function SearchBar:render()
+	-- Render our icon and text box side by side in a Frame
+	return Roact.createElement("Frame", {
+		Size = UDim2.new(0, 200, 0, 20),
+	}, {
+
```
LPGhatguy: Let's delete this line

docs/advanced/refs.md
```diff
@@ -1,7 +1,59 @@
 # Refs
 *Refs* grant access to the actual Instance objects that are created by Roact. They're an escape hatch for when something is difficult or impossible to correctly express with the Roact API.
 
-To create a ref, pass a function prop with the key `Roact.Ref` when creating a primitive element:
+!!! info
+	Refs can only be used with primitive components.
+
+## Refs in Action
+
```
LPGhatguy: We should try to be consistent about whether we have blank lines after each Markdown header. I err on the side of not having a newline, which visually groups sections a little better.

docs/advanced/refs.md
ZoteTheMighty: The newline, yeah?

LPGhatguy: Yep, the empty one!

docs/performance/reduce-reconciliation.md
```diff
@@ -0,0 +1,138 @@
+In all likelihood, the primary source of performance gains for your app will come from reducing the amount of work that Roact's reconcilation process requires. This is accomplished by:
```
LPGhatguy: This page needs a title!

ZoteTheMighty: Technically, the docs will automatically give it the title of the page itself if one isn't present.  But I'll add it for consistency's sake.

PR #85, opened 2018-05-09
Rename reify to mount and teardown to unmount

### LINKED ISSUE #82: Rename 'Roact.teardown' to 'Roact.unmount'
Having two terms for this in the documentation is kind of confusing.

We can introduce a warning for the `teardown` alias but keep it through 1.0.

LPGhatguy:
Closes #82.

This starts to bring Roact's terminology more into focus, and should get rid of a couple of confusing terms: reification and teardown (which doesn't have a good present-tense verb form).

Now, everywhere can refer to mounting and unmounting as _the_ process.

TODO:
* [x] Add warnings to `ReconcilerCompat`
* [x] Update docs
* [x] Improve test coverage for deprecation


LPGhatguy:
I force-pushed my change to hide my mistake as well :)

PR #86, opened 2018-05-10
WIP: Introduce examples and example runner

### LINKED ISSUE #83: Create `examples` folder
It would be useful to introduce an `examples` folder for Roact examples.

It's important that these examples are:
* Easy to run
* Self-contained

LPGhatguy:
Closes #83.

The goal of this PR is to create a simple interactive example system to get people seeing Roact in action much more quickly than the current system. I'm hoping to eventually ship a `RoactWithExamples.rbxlx` with every release.

Some notes:
* I had to remove the benchmarking entries from `rojo.json` because even with 0.4.5 they cause errors
* The example runner is not written in Roact. We may want to rewrite it using Roact before merging.
* Examples need to be updated following the `mount`/`unmount` rename PR.


LPGhatguy:
This should be a very low-risk change, going ahead with a merge.

PR #92, opened 2018-05-15
createRef API

### LINKED ISSUE #70: createRef API like React 16.3
React 16.3+ exposes a [new ref API](https://reactjs.org/docs/refs-and-the-dom.html) that's intended to be the default instead of the callback refs that both React and Roact have.

I've found that a lot of users new to the React model try to do things like call `setState` inside refs, or do all sorts of trickery inside of them, when they really should be doing those things in `didMount` instead.

Switching the recommended API to a `createRef`-like API would prevent these issues in the future.

AmaranthineCodices:
Closes #70.

Adds a `createRef` function to the global Roact namespace. `createRef` returns an object reference with a `current` property that can be used like existing functional refs. Example usage:

```lua
local SomeTextbox = Roact.Component:extend("SomeTextbox")

function SomeTextbox:init()
    self.textboxRef = Roact.createRef()
end

function SomeTextbox:render()
    return Roact.createElement("TextBox", {
        -- ...
        [Roact.Ref] = self.textboxRef
    })
end

function SomeTextbox:didMount()
    self.textboxRef.current:CaptureFocus()
end
```

### DIFF
--- a/docs/advanced/refs.md
+++ b/docs/advanced/refs.md
@@ -8,59 +8,7 @@
 To create a ref, pass a function prop with the key `Roact.Ref` when creating a primitive element.
 
 For example, suppose we wanted to create a search bar that captured cursor focus when any part of it was clicked. We might use a component like this:
-```lua
---[[
-	A search bar with an icon and a text box that captures focus for its TextBox
-	when its icon is clicked
- ]]
-local SearchBar = Roact.Component:extend("SearchBar")
 
-function SearchBar:init()
-	-- Roact.createRef creates an object reference.
-	-- This has a single property, `current`, that can be used to access the
-	-- current Roblox instance.
-	self.textBoxRef = Roact.createRef()
-end
-
-function SearchBar:captureFocus()
-	local textBox = self.textBoxRef.current
-
-	-- If we have a current instance, capture focus on it.
-	-- current will be nil if the component hasn't been mounted yet, or it's
-	-- being unmounted.
-	if textBox then
-		textBox:CaptureFocus()
-	end
-end
-
-function SearchBar:render()
-	-- Render our icon and text box side by side in a Frame
-	return Roact.createElement("Frame", {
-		Size = UDim2.new(0, 200, 0, 20),
-	}, {
-		SearchIcon = Roact.createElement("ImageButton", {
-			Size = UDim2.new(0, 20, 0, 20),
-			-- Handle click events on the icon
-			[Roact.Event.MouseButton1Click] = function()
-				self:captureFocus()
-			end,
-		}),
-
-		SearchTextBox = Roact.createElement("TextBox", {
-			Size = UDim2.new(0, 180, 0, 20),
-			Position = UDim2.new(0, 20, 0, 0),
-			-- We use Roact.Ref to get a reference to the underlying object
-			-- Roact will set textBoxRef.current to the underlying object as
-			-- part of the rendering process.
-			[Roact.Ref] = self.textBoxRef,
-		}),
-	})
-end
-```
-When a user clicks on the outer `ImageButton`, the `captureFocus` method will be called and the `TextBox` instance will get focus as if it had been clicked on directly.
-
-## Functional Refs
-Roact allows you to use functions as refs. The function will be called with the Roblox object that Roact creates. For example, this is the SearchBar component from above, modified to use functional refs instead of object refs:
 ```lua
 --[[
 	A search bar with an icon and a text box that captures focus for its TextBox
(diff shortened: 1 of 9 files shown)


lib/Reconciler.lua
```diff
@@ -93,8 +105,10 @@ function Reconciler.unmount(instanceHandle)
 
 		-- Kill refs before we make changes, since any mutations past this point
 		-- aren't relevant to components.
-		if element.props[Core.Ref] then
-			element.props[Core.Ref](nil)
+		local ref = element.props[Core.Ref]
+
+		if ref then
+			applyRef(ref, nil)
```
LPGhatguy: Let's move this condition into `applyRef` itself, which would make all of these invocations a bit cleaner!

AmaranthineCodices: That'd help clean some stuff up, yeah!

docs/advanced/refs.md
```diff
@@ -8,7 +8,59 @@
...
+			-- part of the rendering process.
+			[Roact.Ref] = self.textBoxRef,
+		}),
+	})
+end
+```
+When a user clicks on the outer `ImageButton`, the `captureFocus` method will be called and the `TextBox` instance will get focus as if it had been clicked on directly.
+
+## Functional Refs
+Roact allows you to use functions as refs. The function will be called with the Roblox object that Roact creates. For example, this is the SearchBar component from above, modified to use functional refs instead of object refs:
```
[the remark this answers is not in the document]
AmaranthineCodices: Fair enough, done!

lib/Reconciler.lua
```diff
@@ -44,6 +44,20 @@ local function isPortal(element)
 	return element.component == Core.Portal
 end
 
+--[[
+	Sets the value of a reference to a new rendered object.
+	Correctly handles both function-style and object-style refs.
+]]
+local function applyRef(ref, newRbx)
+	if not ref then return end
```
LPGhatguy: Let's explicitly check `if ref == nil` and then put the return statement on its own line.

lib/createRef.spec.lua
```diff
@@ -0,0 +1,12 @@
...
+	local createRef = require(script.Parent.createRef)
+
+	it("should create refs", function()
+		expect(createRef()).to.be.ok()
+	end)
+
+	it("should support tostring on refs", function()
+		local ref = createRef()
+		expect(tostring(ref)).to.equal("RoactReference(\"nil\")")
+	end)
```
LPGhatguy: If we're going to test `tostring` on these, let's also verify that they take on the name of their `current` value if set, too!

AmaranthineCodices: That's fair, done!

lib/Reconciler.lua
AmaranthineCodices: Fixed now!

lib/createRef.lua
```diff
@@ -0,0 +1,20 @@
...
+	API is designed to mimic React 16.3's createRef API.
+
+	See:
+	* https://reactjs.org/docs/refs-and-the-dom.html
+	* https://reactjs.org/blog/2018/03/29/react-v-16-3.html#createref-api
+]]
+
+local refMetatable = {
+	__tostring = function(self)
+		return ("RoactReference(%q)"):format(tostring(self.current))
```
LPGhatguy: Thinking about it now, do you think we should just format these with `%s` instead of `%q`? Since they'll mostly be Roblox objects, I don't know if it makes much sense to quote them!

AmaranthineCodices: That's true - I'm used to quoting everything, but in this case it doesn't do much good! Fixed now.

lib/Reconciler.lua
```diff
@@ -323,7 +335,7 @@ function Reconciler._reconcileInternal(instanceHandle, newElement)
 
 		-- Cancel the old ref before we make changes. Apply the new one after.
 		if refChanged and oldRef then
-			oldRef(nil)
+			applyRef(oldRef, nil)
```
LPGhatguy: We need to update the branch on line 349 that applies `newRef` as well! It's calling `newRef` as a function.

lib/Reconciler.spec.lua
```diff
@@ -1,8 +1,39 @@
...
+
+		local element = Core.createElement("StringValue", {
+			[Core.Ref] = ref,
+		})
+
+		local handle = Reconciler.mount(element)
+		expect(currentRbx).to.be.ok()
+		Reconciler.unmount(handle)
+		expect(currentRbx).to.never.be.ok()
+	end)
```
LPGhatguy: Can we add a test that makes sure that refs can be unattached and attached through `Reconciler.reconcile`?

PR #93, opened 2018-05-15
Add stackTrace getter

### LINKED ISSUE #81: Expose element stack traces to components
Components that need to do custom validation work similar to the upcoming prop types system will want to show a stack trace to the element that introduced any bad data.

This is currently accessible via `self._handle._element.source` or something like that, but I'd like to expose a method or property that yields that stack trace if element tracing is enabled.

AmaranthineCodices:
Closes #81. Adds a `stackTrace` method to `Component` that returns the current stack trace for the instance. This will return `nil` when `elementTracing` is `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Added the static lifecycle method `getDerivedStateFromProps` ([#57](https://github.com/Roblox/roact/pull/57))
 * Allow canceling render by returning nil from setState callback ([#64](https://github.com/Roblox/roact/pull/64))
 * Added `defaultProps` value on stateful components to define values for props that aren't specified ([#79](https://github.com/Roblox/roact/pull/79))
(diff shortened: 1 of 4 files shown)


LPGhatguy:
Is there a more descriptive name we can give this? It's the stack trace of where the component's corresponding element was created.

One thing that's fascinating is that this stack trace can _change_ over the lifetime of the component!

AmaranthineCodices:
Yeah, naming is definitely something that can be changed...maybe something like `rootElementTraceback` or `elementTraceback`? I'm not super sure about the exact name, but it's important, since there's two or three possible stack traces a component can have >.<

AmaranthineCodices:
Done!

LPGhatguy:
I keep forgetting this, but can you add an entry to CHANGELOG in the style of the other entries?

PR #94, opened 2018-05-15
validateProps

### LINKED ISSUE #24: Opt-in PropTypes implementation
It would be nice to have a feature similar to React's [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html) functionality.

## Overview
Each stateful component may have a key defined on it - with a symbol, ideally, to avoid name collisions - with a table. This table is structured like this:

```lua
{
    PropertyKey = "Vector3",
    OtherProperty = "table",
    ThirdProperty = Roact.PropTypes.array,
    -- ...
}
```

When the stateful component's `props` table is changed and `DEBUG_ENABLE` has been called, Roact will check each key in the `props` table, match it to the corresponding value in the property types table, and error if the types differ.

The property types need not be limited to simple `typeof(value) == expectedType` expressions. Custom rules that provide more granular type checking allow for checking if, for example, a table is an array (e.g. `ThirdProperty` in the above example).

## Type rules
A "type rule" is the value of an entry in the property types table. It can be one of three possible types:

* A string denoting the expected type as returned by `typeof`.
* An array denoting one or more possible types that the value may be.
* A function that takes a single argument: the value to validate. This function must return `true` or `false`, with an optional second return value explaining why the value was rejected (unused if the function returns `true`).

Some built-in rules will be provided, which will usually be functions. Usually these functions will return the actual validators themselves, e.g. `PropTypes.enumOf("Font")` would return a validator function that returns `true` if and only if the value is an EnumItem from the Font Enum.

## Holes
* Functional components aren't supported. JS allows you to set arbitrary properties of functions; Lua doesn't, so the only way to support functional components would be to call some sort of manual API to validate the props table.
* Kinda on the fence about whether this should even be a part of Roact or just a separate library. It would be nice to have Roact integrate with this regardless, though.

AmaranthineCodices:
Closes #24. This is an integration of [PropTypes](https://github.com/AmaranthineCodices/rbx-prop-types) in Roact.

TODO:

- [x] Gate type checks behind a configuration flag
- [x] Document this

This depends on #93.


LPGhatguy:
I think the prop types implementation, no matter which one we use, should just have a functional interface to maximize the ability to support multiple implementations and custom validators. We probably don't need an `apply` or `validate` method -- just functions would work well!

I imagine the signature for a prop validator function would be `(props, propKey, component) -> success, result`.

Usage would still be:

```lua
Component.propTypes = {
    foo = PropTypes.number, -- or any function of the form above
}
```

AmaranthineCodices:
Some further notes about this:

* Hard dependency on #93 in its current form, where the stack trace method is named `rootElementTraceback`.
* Another stack trace is collected when the component is created; this is used to provide information about what component to look at when `propTypes` is not a function.
* Compatible with any validation function that has the signature `value -> success, reason`.
* Default traceback message has been moved from the reconciler to Core under the name `_defaultSource`. It may need a better name; I'm not sure!

AmaranthineCodices:

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/LPGhatguy/lemur.git
 [submodule "modules/testez"]
 	path = modules/testez
-	url = https://github.com/Roblox/testez.git
+	url = https://github.com/Roblox/testez.git
```
LPGhatguy: Can we revert this change in this PR?

docs/api-reference.md
```diff
@@ -304,4 +304,26 @@ end
...
-	`getDerivedStateFromProps` is a *static* lifecycle method. It does not have access to `self`, and must be a pure function.
\ No newline at end of file
+	`getDerivedStateFromProps` is a *static* lifecycle method. It does not have access to `self`, and must be a pure function.
+
+### validateProps
+```
+static validateProps(props) -> success, reason
+```
+
+Performs property validation. You can use this for type-checking properties; there is a [PropTypes](https://github.com/AmaranthineCodices/rbx-prop-types) library to assist in this.
```
LPGhatguy: We should avoid calling props 'properties' except in their initial introduction. The (shortened) word is used intentionally to explicitly refer to the React and Roact concept as opposed to the general idea of properties.

When we introduce your library, let's say something like "there is a library called *PropTypes* that can be used to validate props declaratively"

lib/Config.lua
```diff
@@ -18,6 +18,8 @@ local defaultConfig = {
 	["elementTracing"] = false,
 	-- Enables instrumentation of shouldUpdate and render methods for Roact components
 	["componentInstrumentation"] = false,
+	-- Enables type checking with the validateProps property on stateful components.
+	["propertyValidation"] = false,
```
LPGhatguy: Much like the recommendation in the docs, let's call this one `propValidation` to be more exact.

lib/Component.lua
```diff
@@ -424,4 +431,33 @@ function Component:_unmount()
...
+			self._extendTraceback),
+		0)
+	end
+
+	local success, failureReason = validator(props)
+
+	if not success then
+		failureReason = failureReason or "<No failure reason was given by the validation function>"
+		error(("Property validation failed:\n%s\n%s"):format(
+			failureReason,
```
LPGhatguy: We should explicitly call `tostring` on the failure reason in case the validator wants to express more detailed errors in some cases.

lib/Core.lua
```diff
@@ -25,6 +25,9 @@ Core.None = Symbol.named("None")
 -- Marker used to specify that the table it is present within is a component.
 Core.Element = Symbol.named("Element")
 
+-- The default "stack traceback" if element tracing is not enabled.
+Core._defaultSource = "\n\t<Use Roact.setGlobalConfig with the 'elementTracing' key to enable detailed tracebacks>\n"
```
LPGhatguy: This could probably stand to have a better name, especially now that it's left the scope of the reconciler! Maybe something like `_elementTracebacksDisabledMessage`?

I'm starting to feel that we should collect some of the error messages together into a module in a future PR. 🤔

lib/Component.lua
```diff
@@ -424,4 +431,33 @@ function Component:_unmount()
 	self._handle = nil
 end
 
+--[[
+	Performs property validation, if it is enabled.
+]]
+function Component:_validateProps(props)
+	if not GlobalConfig.getValue("propertyValidation") then return end
```
LPGhatguy: Let's keep these types blocks to one statement per line, which I think makes them a little easier to read with how keyword-noisy Lua is:

```lua
if not GlobalConfig.getValue("propertyValidation") then
    return
end
```

lib/Component.lua
```diff
@@ -424,4 +431,33 @@ function Component:_unmount()
...
+
+	local validator = self.validateProps
+
+	if validator == nil then return end
+	if typeof(validator) ~= "function" then
+		-- Hide as much as possible about error location, since this message
+		-- occurs from several possible call sites with different stack traces.
+		-- luacheck: ignore 6
+		error(("The value of validateProps must be a function (is a %q). Check the definition of the component extended at:\n%s"):format(
```
LPGhatguy: Can we tweak this error message?

```lua
"The value of validateProps must be a function, but it is a %s." ..
"\nCheck the component defined at:\n%s"
```

Maybe this is a case to bring in `multilineError` or whatever I've been calling it lately!

.gitmodules
AmaranthineCodices: I...can try; this is an artifact of removing `PropTypes`. I'm not sure I'll be able to without borking the branch though. Submodules are weird >.<

lib/Component.spec.lua
```diff
@@ -292,6 +293,106 @@ return function()
...
+			end
+
+			expect(function()
+				Reconciler.mount(Core.createElement(TestComponent))
+			end).to.throw()
+
+			GlobalConfig.reset()
+		end)
+
+		it("should not be run if typeChecking is false", function()
```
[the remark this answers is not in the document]
AmaranthineCodices: Oops, yeah! Fixed now.

lib/Component.lua
```diff
@@ -424,4 +431,35 @@ function Component:_unmount()
...
+	if validator == nil then return end
+	if typeof(validator) ~= "function" then
+		-- Hide as much as possible about error location, since this message
+		-- occurs from several possible call sites with different stack traces.
+		error(("The value of validateProps must be a function, but it is a %s.\n"..
+			"Check the definition of the component extended at:\n%s"):format(
+			typeof(validator),
+			self._extendTraceback),
+		0)
+	end
```
ZoteTheMighty: Is this totally necessary?  We don't bother doing this with lifecycle methods or things like `defaultProps` (making sure it's a table).

I can see the argument for doing it here since you're already opting into validation, just wondering what the thought process is since it's not something we do elsewhere.

AmaranthineCodices: Most of the time when `validateProps` is specified, it's not being written as a method:

```lua
function SomeComponent.validateProps(props)
    -- do some validation work
end
```

Instead, it's usually composed from a bunch of smaller validators, and the result of the composition is assigned. When you're using PropTypes it looks something like this:

```lua
SomeComponent.validateProps = PropTypes.object {
    someKey = PropTypes.number,
    someOtherKey = PropTypes.string,
}
```

The type check is an extra guard against `validateProps` being assigned a non-function, since it's probably not going to be declared as a method of the component class, but as a field of it.

ZoteTheMighty: Gotcha, that makes sense!

LPGhatguy:
An interesting thing that popped into my mind: does gating this behind a global flag make unit testing harder?

Naturally, you want to run all tests with typechecking enabled and also develop with typechecking enabled, but how do you best represent that with the config system only (rightfully?) letting you set configuration once?

AmaranthineCodices:
I'm not entirely sure it blocks unit testing, to be honest, though that may be an artifact of how I run my unit tests. My tests don't run at the same time as the main entry point (they have their own entry point), so the configuration can be set once per entry point without issue.

lib/Component.lua
```diff
@@ -424,4 +431,35 @@ function Component:_unmount()
...
+	Performs property validation, if it is enabled.
+]]
+function Component:_validateProps(props)
+	if not GlobalConfig.getValue("propValidation") then
+		return
+	end
+
+	local validator = self.validateProps
+
+	if validator == nil then return end
```
LPGhatguy: Can we put `return` on its own line? We try to follow one statement per line!

AmaranthineCodices:
Alright! I've gotten this back to mergeable as of now; I think this should be pretty easy to port to the new reconciler!

AmaranthineCodices:
Closing in favor of #174 targeted at the new reconciler.

PR #97, opened 2018-05-19
Add warnings if Name or Parent are specified

### LINKED ISSUE #58: createElement should warn if Name or Parent are specified
`createElement` should give a warning that the `Name` will be unused if specified in properties - it'll always be overwritten with the key in the children table.

The same should be true for Parent; Roact just overwrites the parent during reification, so setting the Parent property in the element props is a performance hit for no gain.

Dog2puppy:
Closes #58. 

A warning will be logged if Name or Parent are specified for createElement(). 

Checklist before submitting:


lib/Core.lua
```diff
@@ -63,6 +63,14 @@ function Core.createElement(elementType, props, children)
 	end
 
 	props = props or {}
+	
+	if props.Name then
+		warn("Name prop is unused. ")
```
AmaranthineCodices: Nitpick: Trailing space in the message.

lib/Core.lua
```diff
@@ -63,6 +63,14 @@ function Core.createElement(elementType, props, children)
...
+	end
+	
+	if props.Parent then
+		warn("Parent prop is unused. ")
```
AmaranthineCodices: Nitpick: Trailing space in the message.

lib/Core.lua
AmaranthineCodices: It'd be nice to explain *why* these props shouldn't be specified, and suggest alternatives. It might also be useful to specify a stack trace so that people can quickly find the source of the warning!

Dog2puppy:
Coveralls doesn’t like it when I use `warn()`. 🙄

lib/Core.lua
Dog2puppy: I have no experience with stacktraces. I’ll try to play around with those.

LPGhatguy:
This change only covers the case where `Name` or `Parent` are specified in the initial mount!

One thing that's tricky about Roact's reconciler right now is that we have to catch cases where state is defined and transitioned separately.

Doing this check in `_setRbxProp` seems like a good idea; we'll also be able to pull the stack trace from the element via `element._source` if it's defined 😄

Dog2puppy:
Never really used Roact yet so I didn’t know about stuff like this. :P Been waiting for 1.0.0.

Dog2puppy:
I think I fixed the conflict stuff.

Dog2puppy:
Should this be merged? 

Also, kinda off topic, but are there any plans of this repo moving with GitHub being purchased by Microsoft?

LPGhatguy:
There's a checklist in your initial post that has an item unchecked! You should look at that!

We definitely need to make sure we're pointing to the right place in the source (ie add stack trace output via `element._source`), and probably need to throw an error instead of just warning.

Make sure you aren't changing any files that aren't part of your PR -- I see some end-of-file line ending changes in `Core.lua` and `CHANGELOG.md`.

Dog2puppy:
I guess I could test using the LogService to see if the warning comes out. 

Already had the `element._source` part, but I don't see why a error would be needed. The original issue mentioned just warning, and since this isn't something that could break everything, don't see you an error would be needed. 

Not sure why there's end of file changes. The change in `CHANGELOG.md` is already there on the master branch. Maybe the GitHub IDE ~~hates me~~ messed it up.

Dog2puppy:
That's 2 issues fixed

Dog2puppy:
Not sure how to fix the error while testing.

LPGhatguy:
I think that setting `Parent` or `Name` is something that should fail tests when you're using Roact, which means that it sort of needs to be an error.

Checking `LogService` for a specific warning seems dubious to me. You'll run (and you have run) into three problems:

* Lemur doesn't have `LogService`, which is what we use to test Roact, so that's a wash
* We'll have random output in the test process because of `warn`, which is something I want to avoid
* Users can't tell if they've done something wrong except by monitoring `LogService` -- I think that these values are pretty critical to know about!

Dog2puppy:
I ended up using a pcall instead. Still not sure why the CI build fails.

ZoteTheMighty:
It looks like the build failed because tests are failing! I'm not actually sure why they're failing, but you also shouldn't need to use pcall here. [TestEZ](https://github.com/Roblox/testez/) has a feature for this:
```lua
-- Function throwing
expect(function()
	error("nope")
end).to.throw()
```

ZoteTheMighty:
Oh, I see! It looks like the reason those don't throw errors is because they're not actually mounting the element.  The reconciler isn't actually trying to set those props until you mount the element that was created.

Dog2puppy:
I saw other tests using `pcalls` for testing for errors. Will fix the testing part so it mounts and uses TestEZ’s function.

Dog2puppy:
15 commits later, I think everything should work now.

Dog2puppy:
Yay, the build passed. Coveralls probably won't like my pull request though.

Dog2puppy:
I think everything is ready to be merged. Might want to add to the docs though.

Dog2puppy:
Can someone please look into this issue?

LPGhatguy:
We're pretty busy and focused on other projects! A warning like this isn't very high priority right now; in practice, very few people hit this case.

We'll probably want the error message to be more informative, something like

```
Name should not be specified as a prop to a primitive component in Roact.
```

If `element._source` is non-nil, which it isn't always (I think your tests fail for a different reason than you expect), you should add:
```
Check call to createElement at:
<stack trace>
```

If not, you should append the `DEFAULT_SOURCE` constant to the message to tell the user how to get a stack trace.

Also, please revert the changes to the files you didn't actual modify, like `Core.lua`, and revert the portion of the change to the `CHANGELOG`.

When you add an entry to `CHANGELOG`, you only need to mention the pull request number, #97, not the issue as well, since any pull requests that have issues will also link to them.

Dog2puppy:
Is there a way to undo a commit with the web view?

LPGhatguy:
This change ended up happening as part of Roact 1.0/the `new-reconciler` branch!

I'm sorry that we weren't responding sooner, but thank you for the PR, it helped us consider fixing this sooner than we otherwise might've.

PR #98, opened 2018-05-21
Change refs in one pass, making switching refs atomic

LPGhatguy:
This is a minor change that ideally shouldn't be user-visible.

Before #92, there was a bug in Roact where when refs were updated, nothing would happen. There was a typo in the reconciler code that resulted in the comparison in the reconciler always determining that refs never changed!

After #92, an issue was unveiled by @ZoteTheMighty via some internal Roact code that means that with the typical use of a ref:

```lua
[Roact.Ref] = (rbx)
   self.foo = rbx
end
```

...there is a period of time where `self.foo` will be `nil` when Roact nullifies the old ref, makes changes, and then passes the same object to the new ref, since by function identity it changed. This doesn't affect code that keeps a stable ref function, and is only visible to code that acts on a ref in response to a change event.

Checklist before submitting:
* [ ] ~Tests added and updated~ I don't think we can test this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Added `defaultProps` value on stateful components to define values for props that aren't specified ([#79](https://github.com/Roblox/roact/pull/79))
 * Added `getElementTraceback` ([#81](https://github.com/Roblox/roact/issues/81), [#93](https://github.com/Roblox/roact/pull/93))
 * Added `createRef` ([#70](https://github.com/Roblox/roact/issues/70), [#92](https://github.com/Roblox/roact/pull/92))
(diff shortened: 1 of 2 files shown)


LPGhatguy:
Ah, based on the small coverage decrease, it seems like we need a test to validate that ref hand-offs happen as we expect, beyond just the fix that this PR makes.

LPGhatguy:
Coverage drop is in... `PureComponent`?

ZoteTheMighty:
Coveralls seems to have gotten very confused about what actually changed...

ZoteTheMighty:
Can confirm that the edge case we saw is resolved by this. Looks good so far, I'll take a look and approve once there's a test that looks good.

PR #101, opened 2018-05-24
Document Roact.None

### LINKED ISSUE #100: Should Document Roact.None

ZoteTheMighty:
Closes #100.

Added information to the API reference indicating how to use Roact.None with setState.

These are just docs changes, so they shouldn't need tests or changelog updates.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -265,6 +265,16 @@ function MyComponent:didMount()
 end
 ```
 
+Since lua tables interpret a `nil` value as absence, removing a value from state is not possible by setting its value to `nil`. If a value needs to be removed from state, it can be set to `Roact.None` when calling `setState`:
+
+```lua
+function MyComponent:didMount()
+	self:setState({
+		keyToRemove = Roact.None
+	})
+end
+```
+
 !!! warning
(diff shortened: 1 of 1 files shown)


docs/api-reference.md
```diff
@@ -265,6 +265,16 @@ function MyComponent:didMount()
 end
 ```
 
+Since lua tables interpret a `nil` value as absence, removing a value from state is not possible by setting its value to `nil`. If a value needs to be removed from state, it can be set to `Roact.None` when calling `setState`:
```
LPGhatguy: Can we tweak this message a little bit?

I want to emphasize that it's *literally impossible* to have a value of `nil` in a table. Having a value of `nil` is exactly the same as not having the value at all!

ZoteTheMighty: How do you feel about:

Note that removing a value from state is not possible by setting its value to `nil` because, in lua tables, `nil` values mean the same thing as no value at all. If a value needs to be removed from state, it can be set to `Roact.None` when calling `setState`, which will ensure that the resulting state no longer contains it:

LPGhatguy: Make sure you capitalize Lua, and then that would be fine I think.

ZoteTheMighty:
Knew I was missing something. I'll keep most of the explanation I have in setState but put some info in Constants as well.

PR #102, opened 2018-05-26
Do not merge: Async retrofit

LPGhatguy:
This is a small experiment to retrofit some asynchronous rendering onto Roact by means of creating (and swapping) implementations of a 'scheduler', a new concept.

The changes are fairly conservative to try to ensure that as much existing code using Roact will still work with this prototype.


LPGhatguy:
This won't ever merge, but it's handy to keep around.

PR #103, opened 2018-05-26
WIP: New reconciler

### LINKED ISSUE #4: New Context API

```lua
function MyComponent:init()
	self._context.blah = "foo"
end
```

```lua

local function UsingValue()
		render = function(theme)
				Text = theme,
			})
		end,
	})

	-- libraries in the past:

			Text = theme,
		})
	end)
end

local function Root()
		value = "overridden value",
	}, {
	})
end
```

### LINKED ISSUE #7: Implement Fragment API
React added the concept of [Fragments](https://reactjs.org/docs/fragments.html), which are elements with multiple children that render into their parent.

I think this would be useful to avoid having junk Roblox instances in the tree, and help flatten some of the gross patterns that have developed with regards to nesting.

### LINKED ISSUE #8: Refactor Portal API to be a regular component
I think Portal can become a regular `Component` with access to protected APIs.

It would make the reconciler much simpler.

### LINKED ISSUE #13: Functional components should have access to context
I'm using the context API to theme components in a hierarchy. At the moment this only works with stateful components, so every component has to be a "stateful" component, even when it actually has no state. React [allows the use of context in functional components](https://reactjs.org/docs/context.html#referencing-context-in-stateless-functional-components) by passing the context table as a parameter. It would be useful to have this functionality, or an equivalent, in Roact as well.

### LINKED ISSUE #18: Asynchronous Rendering
Sometimes, complicated trees result in lots and lots of re-renders.

I'd like to reduce the amount of work that Roact does by batching it and making rendering asynchronous, like React does!

Dan Abramov recently posted a good comment as to why React's `setState` is asynchronous, a change we'll eventually have to make: https://github.com/facebook/react/issues/11527#issuecomment-360199710

### LINKED ISSUE #48: New reconciler
The reconciler is starting to get pretty darn big, complicated, and a bit convoluted.

I'd like to take another stab at rewriting it from the ground up with better-defined types, and to be a little bit more robust.

My wishlist for the new reconciler:
* [Asynchronous Rendering](https://github.com/Roblox/roact/issues/18)
* [Fragments](https://github.com/Roblox/roact/issues/7)
* Better error handling, resolving #55 and #84
* New 'tree root' object, resolving #89 and #90
* Foundation for new context API, from #4
* [Refactored portals](https://github.com/Roblox/roact/issues/8)
* Better `setState` behavior, resolving #56 and #71
* Support for multiple/alternative reconcilers, for out-of-Roblox use and unit testing
* Complete test coverage!

### LINKED ISSUE #84: Errors put Roact in a bad state
The `setState` blocking code can get itself into a bad state if an (unrelated) component causes a bug. This can break everything using Roact if just one failure happens 😱 

There are a couple possible fixes here:
* Introduce an heuristic to let Roact do failure recovery and reset its state
* `pcall` all user methods (this could get expensive 😢)
* Sandbox state like this into a section per disjoint tree
    * This probably isn't a complete solution, but fixes the problem of unrelated code breaking eachother.

### LINKED ISSUE #88: Should Roact warn if a key changes type?
Roact inherits an optimization from React that makes subtrees with the same key reconcile much more quickly. Instead of traversing that children, we just unmount the old tree and replace it with a new one.

The relevant branch of the reconciler: https://github.com/Roblox/roact/blob/ddac5ae8878866799ada06137cd58b5f30a60598/lib/Reconciler.lua#L297-L315

## Problem
This causes some confusing behavior for people who write functions that return components, known as higher-order-components in the React community:

```lua
-- A very contrived component
local function makeFoo(frameProps)
	local Foo = Roact.Component:extend("Foo")

	function Foo:render()
		return Roact.createElement("Frame", frameProps)
	end

	return Foo
end
```

When using makeFoo, the intention is that users would do this:
```lua
local someFoo = makeFoo({
	BorderSizePixel = 0,
})

local function App()
	return Roact.createElement(someFoo)
end
```

However, I've found that users will mistakenly do this:
```lua
local function App()
	return Roact.createElement(makeFoo({
		BorderSizePixel = 0,
	}))
end
```

## Proposed Solution
I think it would be acceptable for Roact to keep this branch, but throw a warning when it's encountered. This would alert a user if they're doing something that causes Roact to be wasteful.

There are some cases where previously I would've used this feature to implement something like a tab panel, where the same key would be reused when the panel changes. However, I think that embracing dynamic keys in that case, by varying the key based on which panel is being displayed, has other benefits for making such a tab panel more flexible.

What gives me hesitation on this:

* Are there any components where reusing keys but changing components is a necessary or superior technique?
* Is a warning the right technique here? Should it be part of dev tooling instead?
* Should this be enabled by default? What about on 'production' builds?

### LINKED ISSUE #99: Roact should mark unmounted components as such for better error messages
If you call `setState` on a component that was unmounted, like if you had a stray changed event, you'll get a really confusing message about how `_handle` wasn't found on some instance deep in Roact's guts.

Roact should _instead_ mark components that it unmounts as being unmounted, and reject any attempts to update them explicitly.

LPGhatguy:
Closes #48, closes #99, closes #88, closes #84, closes #18, closes #13, closes #8, closes #7, closes #4.

This is a *super early* branch that aims to rewrite the entire Roact reconciler as well as the implementation of stateful and functional components.

Checklist before submitting:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,3 +5,4 @@ Closes (ISSUES HERE).
 Checklist before submitting:
 * [ ] Added entry to `CHANGELOG.md`
 * [ ] Added/updated relevant tests
+* [ ] Added/updated documentation


LPGhatguy:
Come on GitHub, that's not... what happened, at all

Issue #47, opened 2018-03-13
Get rid of Core, refactor into several files


LPGhatguy:
I don't think there's any reason to have a giant file called `Core`.

Instead, I'd like to break pretty much everything in that file into its own file and be more selective about what's exported in the public API, which would probably include explicitly laying out each export and getting rid of the giant `apply` function.

LPGhatguy:
1.0 can move on without this, but this is still a good idea.

AmaranthineCodices:
Do you still want to do this, or can I write a PR for it?

LPGhatguy:
I think I'd like to do it. It's not really that interesting of a problem.

AmaranthineCodices:
Okay!

PR #105, opened 2018-05-29
Break createElement and oneChild into their own files (refactor 'Core', part one)

LPGhatguy:
First steps for #47 but doesn't close it yet.

I took `createElement` and `oneChild` out of `Core` and put them into their own files, which makes tests and dependencies much more obvious.

Checklist before submitting:
* [x] Changelog not relevant
* [x] No documentation changes


LPGhatguy:
Coverage increased because the codebase has a couple more lines than it did before!

There are two branches in `createElement` missing coverage, one of which (the warning) is hard to test. I'd like to address those in a future PR.

PR #106, opened 2018-06-04
Errors put Roact in a bad state

TylerRHoyer:
Closes (Errors put Roact in a bad state)

Component level blocks to prevent setState from being called from incorrect lifecycle methods currently are set right before and after a lifecycle call. Instead of blocking these lifecycle methods from calling setState, I recommend making setState private and only passing it into lifecycle methods that should be able to call it. That way when an error occurs there is no component level state that becomes invalid. When an error occurs, the lifecycle method's reference to setState on the lua stack will be removed. To work properly, the reference should not be leaked outside of the lifecycle method's lua stack frame.

Checklist before submitting:


LPGhatguy:
This doesn't really fix the crux of the issue, it only patches over the current symptoms we're seeing.

Another case where errors smash apart the Roact state is in the case of the tree itself. The way to solve this robustly is to `xpcall` all user code, and vigorously typecheck all objects passed by users before they enter into any Roact functions.

The real issue is that it's very, very hard to write code that's exception safe. This is, sadly, a universal problem where I think the best solution is "don't have exceptions" but it's far too late for that. 😄

TylerRHoyer:
Ah yes, the tree. If the reconciler saves its state outside its stack and the entire reconciliation process is pcall'ed at the root, then it will be able to start again, reload its state, and skip the subtree that threw and error. This sounds rigorous to me. It could even use the setState lock method currently implemented.

LPGhatguy:
You can take a look at some of the work I'm doing in #104 with the new reconciler -- there's going to be a new 'task' granularity that serves as a pretty good place to do a lot of this work.

Errors should also only affect individual trees, which means that an error in one Roact tree won't cause everything using Roact to grind to a halt 😄

LPGhatguy:
I'm going to go ahead and close this in favor of something a little bit more comprehensively robust (maybe more `pcall` based on a flag in the new reconciler?)

Good to hear from you, Tyler!

PR #107, opened 2018-06-04
More examples

ZoteTheMighty:
Provide a few more examples in the examples folder.

Checklist before submitting:
* [x] No need for `CHANGELOG.md` entry
* [ ] Not sure about testing needs here
* [x] No need for documentation changes


examples/changed-signal/ReversedText.lua
```diff
@@ -0,0 +1,33 @@
...
+	if not text or text == "" then
+		return result
+	end
+
+	for i = #text, 1, -1 do
+		result = result .. text:sub(i, i)
+	end
+
+	return result
+end
```
LPGhatguy: We should use `string.reverse` instead!

examples/changed-signal/App.lua
```diff
@@ -0,0 +1,43 @@
...
+	self.state = {
+		text = ""
+	}
+end
+
+function App:render()
+	local text = self.state.text
+
+	return Roact.createElement("Frame",{
+			Size = UDim2.new(0, 400, 0, 400),
```
LPGhatguy: This looks a bit double-indented!

examples/changed-signal/App.lua
```diff
@@ -0,0 +1,43 @@
...
+function App:init()
+	self.state = {
+		text = ""
+	}
+end
+
+function App:render()
+	local text = self.state.text
+
+	return Roact.createElement("Frame",{
```
LPGhatguy: Space after comma

examples/changed-signal/App.lua
```diff
@@ -0,0 +1,43 @@
...
+local Roact = require(game.ReplicatedStorage.Roact)
+
+local InputTextBox = require(script.Parent.InputTextBox)
+local ReversedText = require(script.Parent.ReversedText)
+
+local App = Roact.Component:extend("App")
+
+function App:init()
+	self.state = {
+		text = ""
```
LPGhatguy: Trailing comma!

examples/changed-signal/App.lua
```diff
@@ -0,0 +1,43 @@
+--[[
+	Displays a TextBox and a TextLabel that reverses the TextBox's input text
+]]
+local Roact = require(game.ReplicatedStorage.Roact)
+
+local InputTextBox = require(script.Parent.InputTextBox)
+local ReversedText = require(script.Parent.ReversedText)
```
LPGhatguy: Do you think we need this abstraction? It seems like for an example like this, we can probably just inline those objects here, since their transform is so simple. It might make the example a little bit more obvious!

ZoteTheMighty: I was on the fence about this, but I guess we can use bigger examples as places to demonstrate how you might split things into files.

In that regard, should I converge all the examples? The stress test is also simple, but a bit beefier than this one.

ZoteTheMighty: Actually I take it back it's not that bad either.

examples/stress-test/init.lua
```diff
@@ -0,0 +1,85 @@
...
+			self:setState({
+				time = tick(),
+			})
+		end)
+	end
+
+	function App:willUnmount()
+		self.connection:Disconnect()
+	end
+
```
LPGhatguy: Extra newline?

LPGhatguy:
Looks pretty good!

examples/stress-test/init.lua
ZoteTheMighty: Wasn't sure of a better way to separate the component definitions from where I create and mount them. It feels a little weird to have it all inline, but maybe it's just because I'm used to everything all properly compartmentalized into folders.

PR #108, opened 2018-06-04
Docs: highlight createRef over function-style ref

ZoteTheMighty:
Promote the createRef api to be the more favored mode of ref creation, provide further warnings about the function version.

Checklist before submitting:
* [x] No need for `CHANGELOG.md`
* [x] No need for tests
* [x] This IS documentation

### DIFF
--- a/docs/advanced/refs.md
+++ b/docs/advanced/refs.md
@@ -16,6 +16,10 @@ For example, suppose we wanted to create a search bar that captured cursor focus
  ]]
 local SearchBar = Roact.Component:extend("SearchBar")
 
+function SearchBar:init()
+	self.textBoxRef = Roact.createRef()
+end
+
 function SearchBar:render()
 	-- Render our icon and text box side by side in a Frame
 	return Roact.createElement("Frame", {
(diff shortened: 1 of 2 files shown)


docs/api-reference.md
```diff
@@ -110,8 +100,21 @@ function ExampleComponent:didMount()
...
+Roact.createElement("Frame", {
+	-- The provided function will be called whenever the rendered instance changes.
+	[Roact.Ref] = function(rbx)
+		print("Roblox Instance", rbx)
+	end,
+})
+```
+
+!!! warning
+	When `Roact.Ref` is given a funciton, Roact does not guarantee when this function will be run relative to the reconciliation of other props. If you try to read a Roblox property that's being set via a Roact prop, you won't know if you're reading it before or after Roact reconciles that prop!
```
LPGhatguy: In the current reconciler implementation, this isn't actually true. Refs are applied only after all other props are set.

I don't know if we want to make that guarantee, or even how functional refs should have guarantees going forward. I keep thinking more and more that they should be soft-deprecated and eventually removed.

docs/advanced/refs.md
```diff
@@ -16,43 +16,40 @@ For example, suppose we wanted to create a search bar that captured cursor focus
...
+			-- Use Roact.Ref to get a reference to the underlying object
+			[Roact.Ref] = self.textBoxRef
 		}),
 	})
 end
 ```
 When a user clicks on the outer `ImageButton`, the `captureTextboxFocus` callback will be triggered and the `TextBox` instance will get focus as if it had been clicked on directly.
 
-## Refs During Teardown
+## Refs During Unmount
```
LPGhatguy: Good catch!

docs/advanced/refs.md
```diff
@@ -16,43 +16,40 @@ For example, suppose we wanted to create a search bar that captured cursor focus
...
-			-- We use Roact.Ref to get a reference to the underlying object
-			[Roact.Ref] = function(rbx)
 
-				-- Set a callback function to give focus to the TextBox
-				self.captureTextboxFocus = function()
-					rbx:CaptureFocus()
-				end
-			end
+			-- Use Roact.Ref to get a reference to the underlying object
+			[Roact.Ref] = self.textBoxRef
```
LPGhatguy: This example got sooo much simpler.

docs/api-reference.md
ZoteTheMighty: Right, I forgot that this part of the behavior changed. What's odd is that with function refs, you end up with the old one getting called with nil and the new one getting called with the new object. And that'll technically happen every time it re-renders, even if the underlying object doesn't change.

I'm kind of in agreement that I generally want to see the old style of ref go away, because it felt like it was leading towards unsafe uses. What I want here in the docs is some sort of warning that gets at *why* we have negative opinions about it, but I couldn't find a good way to phrase it.

docs/advanced/refs.md
ZoteTheMighty: Oddly enough, it also needed the newline afterwards, otherwise it wasn't rendering at all o.O

docs/api-reference.md
LPGhatguy: We can worry about that soft/hard deprecation later I think, just putting `createRef` first is a big step forward.

LPGhatguy:
LGTM!

PR #111, opened 2018-06-09
Made elements a single table instead of two.

### LINKED ISSUE #109: Remove createElement
On #104 I suggested having createElement remove special markers (like Roact.Children, Roact.Ref, events) from the element props and put them into their own keys in the new element. @LPGhatguy said that there is a beauty to having everything inside of the props. So, taking that idea further, I propose putting the component into the props. This has the following effects: 1, There is no need for the elements struct, meaning that only one table instead of two per element need to be created every render. 2, createElement can be removed, preventing a function call for every element recreated every render. 3, The code becomes much more declarative. 4, We lose the ability to track an element's source to a specific line number (We could still track what render method created it though).

When I tested the performance improvement of this creating an element the new way vs calling createElement, I was surprised to find that it was only an 85% improvement. Running
```lua
for i = 1, 1000000 do
local element = createElement("Part")
end
```
took 5934.6 ms and
```lua
for i = 1, 1000000 do
local element = {
    component = "Part"
}
end
```
took 893.5 ms.

When mounting 10000 parts with only their anchored property set, Roact takes 368 ms more than directly creating the parts. We know that the entire process of element creation is overhead and contributes to the 368 ms total overhead of this operation. The total time createElement took was 17 ms. An 85% reduction in createElement time means a 3.9% reduction in the overhead of instance creation using Roact. This % of overhead reduction would be larger for reconciling because it isn't usually as resource intensive but the createElement call remains at the same intensity. These results don't however account for variable number of properties.

Converting code away from using createElement wouldn't be too difficult because most code could be easily converted using a regular expression. Code that uses a variable as the element properties argument would however need to be manually edited. In addition, an edited version of the createElement function could remain in the code base until all code using it has been changed over to using the new method.

TylerRHoyer:
Partially Closes #109 

While this pull request does not remove createElement or remove element.source, it does move all element properties (component type and source) into the properties, allowing elements to be represented by one table.

Checklist before submitting:

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -244,7 +244,7 @@ end
 	elementTracing configuration flag is set to false.
 ]]
 function Component:getElementTraceback()
-	return self._handle._element.source
+	return self._handle._element[Core.Source]
 end
 
 --[[
--- a/lib/Core.lua
+++ b/lib/Core.lua
@@ -21,4 +21,10 @@ Core.None = Symbol.named("None")
 -- Marker used to specify that the table it is present within is a component.
 Core.Element = Symbol.named("Element")
 
+-- Marker used to specify the component of an element.
+Core.Type = Symbol.named("Type")
+
+-- Marker used to specify stack trace of where an element was created.
+Core.Source = Symbol.named("Element")
+
 return Core
--- a/lib/Instrumentation.lua
+++ b/lib/Instrumentation.lua
@@ -18,6 +18,8 @@
 		- Percent of total shouldUpdate time by component
 ]]
 
+local Core = require(script.Parent.Core)
+
 local Instrumentation = {}
 
 local componentStats = {}
@@ -28,8 +30,8 @@ local componentStats = {}
 ]]
 local function getStatEntry(handle)
 	local name
-	if handle and handle._element and handle._element.component then
-		name = tostring(handle._element.component)
+	if handle and handle._element and handle._element[Core.Type] then
+		name = tostring(handle._element[Core.Type])
 	else
 		warn("Component name not valid for " .. tostring(handle._key))
 		return nil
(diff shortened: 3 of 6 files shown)


LPGhatguy:
This change is pretty wildly breaking for anything that tries to iterate over props, and goes against the grain of what we're trying to do with features like prop type declarations I think.

While there could be a performance argument for allocating half as many tables, I suspect there's a much better way to go about that: improving our allocator, potentially switching it to be generational.

TylerRHoyer:
If the code iterating over the props handles it correctly (ignores all userdata keys like the reconciler does) then they will not have any issues. If their code relies on the internals of the reconciler and was poorly written, then yes, it will break. :P I'm not sure why this would go against the grain of prop type declarations. Could you enlighten me?

LPGhatguy:
Composite components should be able to accept userdata props just fine -- just like they accept (and pass on) other keys like events, refs, and changed connections. There's no reason to special-case a type like this.

The set of code that would break with this change is not the set of bad code. Iterating over all of your properties and validating them is totally sane. Merging properties and meta-information is probably just harmful in this case.

TylerRHoyer:
I understand the concern. The way I see it, elements describe what the reconciler should do to make a component. Given the same element, each component should turn out the exact same (besides stateful components). To accomplish its job, the element stores information which is used in different places in the reconciler. What I find is odd is the structure that the elements store information in. I'd recommend having the element store each piece of information that is handles by the reconciler differently, in different places, or storing all the information in one place. For example, the props table contains information that needs to go to `reconcilePrimitiveProps`, and information that needs to go to `reconcilePrimitiveChildren` as well as `applyRef`, but not the information that gets passed to `Instance.new`. All of these bits of information describe the component, and each piece is needed to render the correct component, but and stored in different places. Now, there is some metadata currently being stored in elements. The `source` and `type` properties are metadata, but the `component` property is not. The `source` can be stored in the global weak key table with the element as the key, and the `type`... well the type system would require some metadata mixed with our data unless we went the global weak keyed table route like I mentioned with `source` metadata.

PR #112, opened 2018-06-09
Refactored reconciler checks

### LINKED ISSUE #110: Refactor component type checks
Right now the isPrimitiveElement, isStatefulElement, and other component type checks are all checking if the element is a table, and returning false if it isn't. This check should be pulled up a scope and done once first. Here is an example of how I think it should be done:

```lua
if type(element) == "table" then
    local componentType = type(element.component)
    if componentType == "table" then -- Stateful Component
        …
        return
    elseif componentType == "function" then -- Functional Component
        …
        return
    elseif …
        …
        return
    end
end
error(("Cannot mount invalid Roact element %q"):format(tostring(element)))
```

This does a couple things. First, it prevents checking if an element is a table multiple times. Second, it makes the checks inline rather than multiple function calls. Third, it prevents each check from calling type multiple times by caching the result. What are your thoughts on this?

TylerRHoyer:
Closes #110 .

Refactored reconciler checks to be more efficient as outlined in #110.

Checklist before submitting:

### DIFF
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -62,9 +62,9 @@ Reconciler._singleEventManager = SingleEventManager.new()
 ]]
 function Reconciler.unmount(instanceHandle)
 	local element = instanceHandle._element
-	local elementType = type(element.component)
+	local componentType = type(element.component)
 
-	if elementType == "string" then
+	if componentType == "string" then
 		-- We're destroying a Roblox Instance-based object
 
 		-- Kill refs before we make changes, since any mutations past this point
@@ -79,12 +79,12 @@ function Reconciler.unmount(instanceHandle)
 		Reconciler._singleEventManager:disconnectAll(instanceHandle._rbx)
 
 		instanceHandle._rbx:Destroy()
-	elseif elementType == "function" then
+	elseif componentType == "function" then
 		-- Functional components can return nil
 		if instanceHandle._child then
 			Reconciler.unmount(instanceHandle._child)
 		end
-	elseif elementType == "table" then
+	elseif componentType == "table" then
 		instanceHandle._instance:_unmount()
 	elseif element.component == Core.Portal then
 		for _, child in pairs(instanceHandle._children) do
@@ -100,22 +100,6 @@ end
 	component tree.
 ]]
 function Reconciler.mount(element, parent, key)
-	if type(element) ~= "table" then
-		if type(element) == "boolean" then
-			-- Ignore booleans of either value
-			-- See https://github.com/Roblox/roact/issues/14
-			return nil
-		end
-
-		local message = (
-			"Bad argument #1 to Reconciler.mount, expected a table representing an element, found a %s"
-		):format(
-			typeof(element)
-		)
-
-		error(message, 2)
-	end
-
 	return Reconciler._mountInternal(element, parent, key)
 end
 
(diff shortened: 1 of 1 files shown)


lib/Reconciler.lua
```diff
@@ -64,75 +56,47 @@ local Reconciler = {}
...
 --[[
 	Destroy the given Roact instance, all of its descendants, and associated
 	Roblox instances owned by the components.
 ]]
 function Reconciler.unmount(instanceHandle)
 	local element = instanceHandle._element
 
-	if isPrimitiveElement(element) then
-		-- We're destroying a Roblox Instance-based object
+	if type(element) == "table" then
```
LPGhatguy: Instead of indenting the entire function one level, we should probably just early-out with an error if `element` isn't a table.

That being said, in this function, `element` should _always_ be a table, so this check is just redundant.

lib/Reconciler.lua
```diff
@@ -156,115 +120,118 @@ end
 	the reconciliation methods; they depend on this structure being well-formed.
 ]]
 function Reconciler._mountInternal(element, parent, key, context)
-	if isPrimitiveElement(element) then
-		-- Primitive elements are backed directly by Roblox Instances.
+	if type(element) == "table" then
```
LPGhatguy: Same here, probably not worth checking if the element is a table; we can do typechecking assertions in `mount` though in a later PR, since it's a user-facing API.

TylerRHoyer:
I've removed some checks, and moved others to the public version of the method. Also, changed the logic for the logic in the public version of the reconcile method. It was checking if the instanceHandle was nil then indexing it. It would throw a nasty error if instanceHandle wasn't a table, so I made it first check if the instanceHandle is a table then check if it has the isInstanceHandle symbol.

LPGhatguy:
If possible, I'd like to keep the scope of PRs like this pretty small.

Can we scale this one back to _just_ refactoring the checks on `element.component`? That's a change I would be cool with merging sooner, but reviewing some of the other things here is more difficult.

TylerRHoyer:
I moved error in unmounts back into the elseif statement. Unfortunately refactoring the `element.component` checks means also moving the `type(element)` checks, and if we move them then the best spot to put them is out of the internal functions, as you mentioned. I'm not sure what else could be removed from this commit, do you see anything?

lib/Reconciler.lua
```diff
@@ -140,6 +100,22 @@ end
 	component tree.
 ]]
 function Reconciler.mount(element, parent, key)
+	if type(element) ~= "table" then
+		if type(element) == "boolean" then
+			-- Ignore booleans of either value
+			-- See https://github.com/Roblox/roact/issues/14
+			return nil
+		end
```
LPGhatguy: Pretty sure moving this check here will break lots of code, since everything in the reconciler uses `_mountInternal`, not `mount.`

lib/Reconciler.lua
```diff
@@ -264,10 +241,6 @@ function Reconciler._mountInternal(element, parent, key, context)
 			_children = children,
 			_rbx = target,
 		}
-	elseif typeof(element) == "boolean" then
-		-- Ignore booleans of either value
-		-- See https://github.com/Roblox/roact/issues/14
-		return nil
```
LPGhatguy: We should keep this check inside this function. If we're going to move it, let's put it first so that we can avoid indexing into a bool/nil element.

LPGhatguy:
This has given me some good ideas for how I think we want to check for components in the future.

I'm afraid that just flat-out removing the current `isFoo` checks causes the code to be a little less readable.

I started up a new branch named `element-kind-checks` that will have a PR open soon that takes a slightly different approach. Instead of having a bunch of `isWhatever` checks, I added a `getElementKind` function that classifies an element in one go.

LPGhatguy:
I think I'm going to close this in favor of #123. Thanks for the work on this, it was super useful in figuring out where to take this sort of code I think.

PR #113, opened 2018-06-09
Edited misleading name comment in reconciler

TylerRHoyer:
Changed comment amount names not affecting anything and being used solely for debugging to mention how the reconciler assumes elements with the same name/key between reconciles are old/new versions of the same element.

Checklist before submitting:


LPGhatguy:
Nice!

PR #114, opened 2018-06-09
Pcall once per props instead of prop

TylerRHoyer:
Refactored the _reconcilePrimitiveProps, it was using a table called 'seenProps' when it didn't need to be. Also made it submit batches of props to change in a call to _setRbxProps (previously _setRbxProp). _mountInternal now also uses _setRbxProps. _setRbxProp shouldn't ever error, but will warn if setting a property wasn't successful and continue onto the next properties. Not only did I run the tests, but I made a couple custom tests myself and verified that the output looks pretty.

Checklist before submitting:

### DIFF
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -162,9 +162,7 @@ function Reconciler._mountInternal(element, parent, key, context)
 		local rbx = Instance.new(element.component)
 
 		-- Update Roblox properties
-		for key, value in pairs(element.props) do
-			Reconciler._setRbxProp(rbx, key, value, element)
-		end
+		Reconciler._setRbxProps(rbx, element.props, element.source)
 
 		-- Create children!
 		local children = {}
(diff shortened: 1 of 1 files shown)


vim.exe.stackdump
```diff
@@ -0,0 +1,19 @@
...
+00000000001  00180135376 (0010065D540, 00000000008, 00000000000, 00000000000)
+00000000001  0018011C6FB (0010065D540, 00000000008, 00000000000, 00000000000)
+00000000001  001004F77E4 (00100575E0A, 00000000000, 00000000001, 00100663D0C)
+00000000001  001005820F3 (00000000008, 00100664140, 00000000000, 00000000000)
+00000000001  00100576D86 (00000000000, 0005ADAE9CE, 00024652830, 00000010000)
+00000000001  001005D4CEB (000FFFFCCB0, 00000000000, 0018011C6FB, 00000000001)
+00600049190  001005E218A (00000000020, 001802E0900, 00180047986, 00180046990)
+000FFFFCCB0  001800479F7 (00000000000, 00000000000, 00000000000, 00000000000)
+00000000000  00180045663 (00000000000, 00000000000, 00000000000, 00000000000)
+End of stack trace (more stack frames may be present)
```
LPGhatguy: Nice.

TylerRHoyer: lol, I guess "git stage ." is a little dangerous!

TylerRHoyer: I've amended the commit

TylerRHoyer:
This PR is ready for a review.

LPGhatguy:
RDC is over! I have time for PRs!

Okay so -- I think the premise of this PR is good. Reducing the number of `pcall` operations we have is handy from an error handling perspective and potentially a performance perspective too.

However, I think this change is aiming to do a little too much in one pass. Additionally, I'm not all that interested in trying to land refactors (besides feature backports) of the current reconciler right now.

LPGhatguy:
Circling back around, I think I'm going to close this PR in favor of smaller changes to the existing reconciler, as well as doing these sorts of changes in the new reconciler instead.

PR #116, opened 2018-06-11
Resolved PureComponent render issue when using defaultProps

zoebasil:
Fixed bug with defaultProps not being included in newProps when passed into component's shouldUpdate method, causing components to rerender when unnecessary

### DIFF
--- a/lib/Component.spec.lua
+++ b/lib/Component.spec.lua
@@ -228,11 +228,6 @@ return function()
 			bar = "world",
 		}
 
-		function TestComponent:shouldUpdate(newProps)
-			lastProps = newProps
-			return true
-		end
-
 		function TestComponent:render()
 			lastProps = self.props
 			return nil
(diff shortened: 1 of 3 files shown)


LPGhatguy:
The way Git generated a diff here was _super_ confusing!

Can you add a test case to `Component` that verifies that values from `defaultProps` are correctly passed into `shouldUpdate`?

lib/Component.spec.lua
```diff
@@ -262,6 +267,19 @@ return function()
...
+		Reconciler.reconcile(handle, createElement(TestComponent, {
+			baz = "!",
+		}))
+
+		expect(lastProps).to.be.a("table")
+		expect(lastProps.foo).to.equal("hello")
+		expect(lastProps.bar).to.equal("world")
+		expect(lastProps.baz).to.equal("!")
+
+		Reconciler.unmount(handle)
```
ZoteTheMighty: Does this test guarantee that `shouldUpdate` got the new props?  `lastProps` gets set by both `shouldUpdate` and `render`, so `render` will be the last one to have modified this variable.

It may be a good idea to break this into a new test.

zoebasil: oh you're right, i didn't think about that. i'll split it into a separate case

zoebasil: sure (:

LPGhatguy:
Looks good! Another case popped up with `init` that might be a related bug. I'll merge and then take a look at that case too.

PR #117, opened 2018-06-11
Advanced error handling

TylerRHoyer:
Closes (#84 ).

It's a good thing no code is sacred!

This PR adds my "Heapstack" class which sets up a stack on the heap. It stores function call parameters on the heap and calls one function at a time, only moving onto another function when the running function returns. Running functions can push another function to run on the stack, the last function pushed is the first run once the function returns. Since functions are not being called inside each other, little actual lua stack space is used, and only the currently running function's frame is removed when an error occurs. Since stack traces are no longer helpful, I've implemented Heapstack traces which can be enabled when a Heapstack is created. Note that functions can still use the stack and make nested calls, but they wont be safe from errors. Also, Heapstack traces don't include calls allocated on the stack, allowing some gaps in the trace to occur if the Heapstack isn't used religiously. (While I believe this could be fixed, it doesn't seem worth it to me.) 

I've included the changes from my previous PR, "Enhanced error handling", and "Refactored checks". This is a WIP and I still have the following to work on: 9 failing tests, look into integrating the _setRbxProps enhanced error handling with Heapstack (they look similar), lots of code cleanup, and running benchmarks. Unfortunately I don't see these changes positively affecting the benchmarks, but what error handling does? Hopefully the effect will be minimal and optimizations can be made. I don't think that we need to be using the Heapstack everywhere, only where we need to protect the state. That being said, I did use it pretty much everywhere.

I found myself getting really annoyed with all those pesky, "expect ___ to throw" test cases and may have removed one to two of them since Roact shouldn't be throwing errors anymore.

What are your thoughts on this?

Checklist before submitting:

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -19,6 +19,7 @@ local Reconciler = require(script.Parent.Reconciler)
 local Core = require(script.Parent.Core)
 local GlobalConfig = require(script.Parent.GlobalConfig)
 local Instrumentation = require(script.Parent.Instrumentation)
+local Heapstack = require(script.Parent.Heapstack)
 
 local invalidSetStateMessages = require(script.Parent.invalidSetStateMessages)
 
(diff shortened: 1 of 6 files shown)


LPGhatguy:
This is nuts! We're not doing to switch to a different mechanism of calling functions and ditch Lua's regular stack for a table-backed 'heap stack'.

TylerRHoyer:
Lol! Perhaps I have scared you, but consider it a simple task scheduler that just runs tasks LIFO. It isn't all to different from the system in the new reconciler. Also, to recover from errors, the current solution without creating "tasks" in a task scheduler or "frames" in a heapstack would be to use a different calling mechanism, pcall.

BTW, I'm working a nightshift which accounts for my night activity on GitHub.

PR #132, opened 2018-07-19
Refactor SingleEventManager

LPGhatguy:
This PR should make SingleEventManager a little bit tidier. Instead of indexing by string keys, the internal event table is indexed by the event objects themselves.

Additionally, this refactored `SingleEventManager` uses an indirection trick to make replacing an existing connection much quicker. When connecting to an event that's already connected, we can just update a member in a table instead of disconnect/reconnecting a signal.

This PR relies on the idea that `GetPropertyChangedSignal` will always return the same signal every time it's called, since we use it for object identity.

Checklist before submitting:
* [x] Not user-facing change

### DIFF
--- a/lib/SingleEventManager.lua
+++ b/lib/SingleEventManager.lua
@@ -20,36 +20,23 @@ local SingleEventManager = {}
 
 SingleEventManager.__index = SingleEventManager
 
-local function createHook(rbx, key, method)
+local function createHook(rbx, event, method)
 	local hook = {
 		method = method,
-		connection = rbx[key]:Connect(function(...)
-			method(rbx, ...)
-		end)
 	}
 
-	return hook
-end
-
-local function createChangeHook(rbx, key, method)
-	local hook = {
-		method = method,
-		connection = rbx:GetPropertyChangedSignal(key):Connect(function(...)
-			method(rbx, ...)
-		end)
-	}
+	hook.connection = event:Connect(function(...)
+		hook.method(rbx, ...)
+	end)
 
 	return hook
-end
-
-local function formatChangeKey(key)
-	return ("!PropertyChangeEvent:%s"):format(key)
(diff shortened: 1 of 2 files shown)


lib/SingleEventManager.lua
```diff
@@ -14,132 +14,145 @@
...
 
-local function formatChangeKey(key)
-	return ("!PropertyChangeEvent:%s"):format(key)
+	return self
 end
 
-function SingleEventManager.new()
-	local self = {}
+function SingleEventManager:connect(instance, key, method)
+	self:_connectInternal(instance, instance[key], method)
```
AmaranthineCodices: Do we want to try to check that `key` is pointing to an event here, and throw a nice error?

lib/SingleEventManager.lua
```diff
@@ -14,132 +14,145 @@
...
+		self._hooks[instance] = instanceHooks
+	end
 
-			existingHook.connection:Disconnect()
-		end
+	local existingHook = instanceHooks[event]
 
-		rbxHooks[formattedKey] = createChangeHook(rbx, key, method)
+	if existingHook ~= nil then
+		existingHook.method = method
```
AmaranthineCodices: This is *nice*. I wonder how much this speeds up changing event connections in the case where they're defined in `render`?

LPGhatguy: When I did some unscientific napkin benchmarking, it's somewhere around 100x faster in the worst case. Not sure if it was ever a bottleneck, but it's interesting!

lib/SingleEventManager.lua
LPGhatguy: Hm! I think that would be kind of problematic in this case, since we don't have something like `FindFirstChild` to use as a non-throwing variant.

PR #134, opened 2018-07-24
Do Not Merge: Hack in bindings onto the reconciler

LPGhatguy:
This is an experiment to tinker with the ergonomics of introducing "bindings" -- special values that can be used with primitive elements to bypass the reconciler.

@cwlowder and I were curious to see if any problems pop up purely in their ergonomics, so I hacked them into the reconciler with the minimal set of code changes required.

The interface Roact expects for a binding in this PR is a table with:

* `current`: The current value the binding is set to
* `subscribe`: A function that adds a listener to the binding, returning a disconnect function

As a hack, I added a new member to instance handles named `_bindings` that keeps track of each binding by key. Transitions to and from bindings _should_ work correctly, but I haven't tested any of this code.

### DIFF
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -165,9 +165,22 @@ function Reconciler._mountInternal(element, parent, key, context)
 
 		local rbx = Instance.new(element.component)
 
+		local bindings = {}
+
 		-- Update Roblox properties
 		for key, value in pairs(element.props) do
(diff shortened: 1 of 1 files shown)


LPGhatguy:
Uhh, thanks, Coveralls.

LPGhatguy:
I think we can move on from this now that we have real, non-hack attempts at this same problem!

PR #139, opened 2018-07-30
Allow setState inside willUpdate

LPGhatguy:
This change lays a little bit of groundwork for allowing `setState` in more places.

Right now, `setState` is disallowed in all lifecycle hooks and `render` with the exception of `didMount`. I think it's safe to expand that to `willUpdate` as well, predicated on the behavior of `setState` not causing a re-render at that point, but instead updating the existing state.

This PR adds `_setStateWithoutUpdate` next to `_setStateBlockedReason`, which changes the behavior of a state update.

Checklist before submitting:

### DIFF
--- a/lib/Component.lua
+++ b/lib/Component.lua
@@ -92,6 +92,11 @@ function Component:extend(name)
 		-- You can see a list of reasons in invalidSetStateMessages.
 		self._setStateBlockedReason = nil
 
+		-- When set to true, setState should not trigger an update, but should
+		-- instead just update self.state. Lifecycle events like `willUpdate`
+		-- can set this to change the behavior of setState slightly.
+		self._setStateWithoutUpdate = false
+
 		if class.defaultProps == nil then
 			self.props = props
 		else
@@ -236,7 +241,12 @@ function Component:setState(partialState)
 	end
 
 	local newState = merge(self.state, partialState)
-	self:_update(nil, newState)
+
+	if self._setStateWithoutUpdate then
+		self.state = newState
+	else
+		self:_update(nil, newState)
+	end
 end
 
 --[[
@@ -314,9 +324,9 @@ end
 ]]
 function Component:_forceUpdate(newProps, newState)
 	if self.willUpdate then
-		self._setStateBlockedReason = "willUpdate"
+		self._setStateWithoutUpdate = true
 		self:willUpdate(newProps or self.props, newState or self.state)
-		self._setStateBlockedReason = nil
+		self._setStateWithoutUpdate = false
 	end
 
 	local oldProps = self.props


LPGhatguy:
This probably isn't quite ready (tests, docs, and changelog entry needed still), but probably ready for someone to see if this is sound.

LPGhatguy:
The original intent with locking down `setState` in all of our lifecycle methods was mostly trying to be as conservative as possible until we figure out the way to go.

React handles `setState` differently based on when in the pipeline it's called. When it's called when there's already a render scheduled, it just updates the state that the next render will use. That lines up with what I wanted to do with the new (async) reconciler, though the implementation will be drastically different there.

ZoteTheMighty:
Okay, I think I see where you're coming from.  It definitely would be good to have fewer situations where users have to worry about creating infinite re-rendering scenarios, and I suspect `setState` in `willUpdate` provides a more natural way to do the sorts of things you'd do in `getDerivedStateFromProps`.

This looks fine so far, I'll look again and review when you tick some boxes.

docs/api-reference.md
```diff
@@ -203,7 +203,18 @@ init(initialProps) -> void
...
+```lua
+function MyComponent:init()
+	self:setState({
+		position = 0,
+		velocity = 10
+	})
+end
+```
+
+In older versions of Roact, `setState` was disallowed in `init`, and you would instead assign to `state` directly. This is still acceptable, but it's simpler to use `setState`:
```
ZoteTheMighty: Wording right before the colon here is a bit confusing as it sounds like you're about to demonstrate use of `setState`. Could you construct the sentence differently? Maybe "It's simpler to use `setState`, but it's still acceptable to assign it directly:"

docs/api-reference.md
```diff
@@ -281,6 +292,16 @@ end
 
 Setting a field in the state to `Roact.None` will clear it from the state. This is the only way to remove a field from a component's state!
 
+!!! warning
+	`setState` can be called from anywhere **except**:
+
+	* Lifecycle hooks: `willUnmount`
+	* Pure functions: `render`, `shouldUpdate`
+
+	Calling `setState` inside of `init` or `willUpdate` has different behavior in other places. Because Roact is already going to render or update a component in these cases, the update will be replaced instead of another update happening.
```
ZoteTheMighty: I don't like the "in other places" part (also, you're missing a "than").  It might be fine to just say "...has special behavior." What do you think?

docs/guide/state-and-lifecycle.md
```diff
@@ -3,15 +3,48 @@ In the previous section, we talked about using components to create reusable chu
...
+		return {
+			currentTime = currentTime + state.currentTime
+		}
+	end)
+end
+```
+
+In this case, we're passing a _function_ to `setState`. This function is called and passed the current state, and returns a new state. It can also return `nil` to abort the state update, which lets Roact make some handy optimizations.
+
+Right now, this version of `setState` is exactly the same as the form where we pass an object. In the future, Roact will support a number of features, like asynchronous rendering, that make the distinction more important.
```
ZoteTheMighty: I'd maybe say "**works** exactly the same"

lib/Component.spec.lua
```diff
@@ -455,31 +458,40 @@ return function()
 
 			local testElement = createElement(TestComponent)
 
-			expect(function()
-				Reconciler.mount(testElement)
-				forceUpdate()
-			end).to.throw()
+			local handle = Reconciler.mount(testElement)
+
+			expect(forceUpdate).to.be.a("function")
```
ZoteTheMighty: Not a problem, but I'm curious why you added this.

LPGhatguy: My thought process was that this expectation is sort of implicit, and if it ever failed, the error message would be not-so-great. Thinking about it now, I think I'll remove it.

PR #145, opened 2018-08-07
WIP: Implement event suspension to try to make listener behavior more intuitive

LPGhatguy:
I don't know if this is the right approach, but I'd like to commit this to a branch and see if it sticks.

The idea here is that while Roact is in the 'commit' phase (actually mutating elements) we should suspend listeners until we get out of the commit phase. The original goal was to reduce the amount of times we have to call `spawn` or `delay` inside of changed listeners that want to call `setState`, but now that I've implemented it, I see that this doesn't help.

Instead, perhaps the best approach is to make `setState` specifically not apply until after the commit phase is over. That way, we have knowledge at the level of the stateful component instead of trying to infer some lifecycle from inside the commit method of a primitive/host component.

Checklist before submitting:

### DIFF
--- a/lib/Reconciler.lua
+++ b/lib/Reconciler.lua
@@ -166,6 +166,8 @@ function Reconciler._mountInternal(element, parent, key, context)
 
 		local rbx = Instance.new(element.component)
 
+		Reconciler._singleEventManager:suspendEvents(rbx)
+
 		-- Update Roblox properties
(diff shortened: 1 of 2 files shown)


LPGhatguy:
I think it's safe to close this in favor of pushing for the new reconciler more aggressively.

PR #148, opened 2018-08-14
Fix cases where properties don't correctly reconcile to nil

### LINKED ISSUE #147: Reconciling from an event to nil causes weird errors
This section of code shouldn't run if the key isn't a string:

https://github.com/Roblox/roact/blob/9bb4fbaf6bd601cdbd3a7b93e4a3f8c29e9cfbbd/lib/Reconciler.lua#L460-L469

LPGhatguy:
Fixes #147.

This fixes a couple cases based on reconciling an event or change handler to `nil`.

The issues were:
* `getDefaultPropertyValue` was invoked before we classified the property
* The error value from `getDefaultPropertyValue` was ignored, meaning an error message would be set to the default value when reconciling to `nil` for invalid keys like symbols
* The reconciler never handled reconciling to `nil` values for event and change handlers and thus never called `SingleEventManager:disconnect`.

Checklist before submitting:
* ~Added/updated documentation~


lib/Reconciler.spec.lua
```diff
@@ -85,4 +89,62 @@ return function()
...
+			if property == "Size" then
+				sizeChangedCount = sizeChangedCount + 1
+			end
+		end
+
+		local element = createElement("Frame", {
+			[Event.Changed] = eventCallback,
+		})
+
+		local handle = Reconciler.mount(element, ReplicatedStorage, "Foo")
```
LPGhatguy: Let's mount into a container that we create in the test so that we don't leave cruft in the DataModel in the case the test fails.

LPGhatguy:
Looks good to me, since I can't formally approve my own pull request apparently.

PR #150, opened 2018-08-21
WIP: New reconciler stuff

AmaranthineCodices:
Expanding the unit test suite for the new reconciler, plus some attempts at comments and bugfixes.

TODO:
- [x] `strict`
- [x] `ChildUtils`
- [x] `Type`
- [x] `createSpy`
- [x] expand `createReconciler`
- [x] expand `RobloxRenderer`

### DIFF
--- a/lib/ChildUtils.lua
+++ b/lib/ChildUtils.lua
@@ -6,8 +6,35 @@ end
 
 local ChildUtils = {}
 
+--[[
+	A signal value indicating that a child should use its parent's key, because
+	it has no key of its own.
+
+	This occurs when you return only one element from a functional component or
+	stateful render function.
+]]
 ChildUtils.UseParentKey = {}
 
+--[[
+	Returns an iterator over the children of an element.
+	`childrenOrChild` may be one of:
+	* a boolean
+	* nil
+	* a single element
+	* a table of elements
+
+	If `childrenOrChild` is a boolean or nil, this will return an iterator with
+	zero elements.
+
+	If `childrenOrChild` is a single element, this will return an iterator with
+	one element: a tuple where the first value is ChildUtils.UseParentKey, and
+	the second is the value of `childrenOrChild`.
+
+	If `childrenOrChild` is a table, this will return an iterator over all the
+	elements of the array, equivalent to `pairs(childrenOrChild)`.
+
+	If `childrenOrChild` is none of the above, this function will throw.
+]]
 function ChildUtils.iterateChildren(childrenOrChild)
 	local richType = Type.of(childrenOrChild)
 
@@ -40,6 +67,15 @@ function ChildUtils.iterateChildren(childrenOrChild)
 	error("Invalid children")
 end
 
+--[[
+	Gets the child corresponding to a given key, respecting Roact's rules for
+	children. Specifically:
+	* If `elements` is nil or a boolean, this will return `nil`, regardless of
+	  the key given.
+	* If `elements` is a single element, this will return `nil`, unless the key
+	  is ChildUtils.UseParentKey.
+	* If `elements` is a table of elements, this will return `elements[key]`.
+]]
 function ChildUtils.getChildByKey(elements, key)
 	if elements == nil or typeof(elements) == "boolean" then
 		return nil
--- a/lib/ChildUtils.lua
+++ b/lib/ChildUtils.lua
@@ -10,7 +10,7 @@ local ChildUtils = {}
 	A signal value indicating that a child should use its parent's key, because
 	it has no key of its own.
 
-	This occurs when you return only one element from a functional component or
+	This occurs when you return only one element from a function component or
 	stateful render function.
 ]]
 ChildUtils.UseParentKey = {}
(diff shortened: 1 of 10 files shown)


lib/ChildUtils.spec.lua
```diff
@@ -0,0 +1,84 @@
+return function()
+	local ChildUtils = require(script.Parent.ChildUtils)
+	local createElement = require(script.Parent.createElement)
```
LPGhatguy: Can we get away with using dummy tables as children instead of elements?

I think I updated `iterateChildren` to no longer care about the type of the children since this function would also be useful for iterating over the children of nodes if we go that route.

AmaranthineCodices: The blocker to that when I was initially writing those tests was that we use the rich type in `iterateChildren` to determine if we were passed an element, or a table of elements - they're both tables in the eyes of `typeof`:

https://github.com/Roblox/roact/blob/ef681f3ce2a45dd68f96d38023daf3051e6e2125/lib/ChildUtils.lua#L39

We could just falsify the type, but that does feel a little weird.

LPGhatguy: Hm, I guess that's true, this seems fine then.

lib/ChildUtils.lua
```diff
@@ -6,8 +6,35 @@ end
 
 local ChildUtils = {}
 
+--[[
+	A signal value indicating that a child should use its parent's key, because
+	it has no key of its own.
+
+	This occurs when you return only one element from a functional component or
```
LPGhatguy: Nit: functional components are becoming just 'function components' to differentiate them from... components that aren't broken.

lib/ChildUtils.spec.lua
```diff
@@ -0,0 +1,84 @@
...
+			local children = {
+				a = createElement("TextLabel"),
+				b = createElement("TextLabel"),
+			}
+
+			local count = 0
+
+			for key, child in ChildUtils.iterateChildren(children) do
+				expect(typeof(key)).to.equal("string")
+				expect(Type.of(child)).to.equal(Type.Element)
```
LPGhatguy: We should check that we're getting the right children, too!

lib/ChildUtils.lua
AmaranthineCodices: Oops, yeah, I missed that!

lib/ChildUtils.spec.lua
AmaranthineCodices: Done! It's kind of troublesome because of no guarantee around iteration order x.x

lib/ChildUtils.spec.lua
```diff
@@ -0,0 +1,88 @@
...
+
+		it("should return the corresponding element", function()
+			local children = {
+				a = createElement("TextLabel"),
+				b = createElement("TextLabel"),
+			}
+
+			expect(ChildUtils.getChildByKey(children, "a")).to.equal(children.a)
+			expect(ChildUtils.getChildByKey(children, "b")).to.equal(children.b)
+		end)
```
LPGhatguy: Can we test each case where we ask for an invalid key? We want to make sure that asking for `c` doesn't give us anything.

AmaranthineCodices: Done now!

AmaranthineCodices:
I think this is in a pretty good spot to merge now.

lib/RobloxRenderer.spec.lua
```diff
@@ -73,6 +74,77 @@ return function()
...
+			local _, defaultStringValue = getDefaultPropertyValue("StringValue", "Value")
+
+			local element = createElement("StringValue", {
+				Value = firstValue
+			}, {
+				ChildA = createElement("IntValue", {
+					Value = 1
+				}),
+				ChildB = createElement("BoolValue", {
+					Value = false,
```
LPGhatguy: We should set this value to `true` so that when we reconcile over the top, we're actually changing the value!

lib/RobloxRenderer.spec.lua
```diff
@@ -73,6 +74,77 @@ return function()
...
+				ChildD = createElement("StringValue", {
+					Value = "test",
+				})
+			})
+
+			local node = reconciler.createNode(element, parent, key)
+			RobloxRenderer.mountHostNode(reconciler, node)
+
+			-- Not testing mountHostNode's work here, only testing that the
+			-- node is properly updated.
```
LPGhatguy: This brings up an interesting general testing philosophy. I like to assert that the construction was correct here anyways, since it's possible that a bug in mounting causes this test to pass mistakenly!

At the same time, there is an argument for only testing the thing you're actually trying to test. I'm not entirely sure!

AmaranthineCodices: I wanted to cut down on clutter in this test - it's already pretty long, and verifying that mounting worked wouldn't provide any benefits on top of the existing `mountHostNode` tests.

lib/RobloxRenderer.spec.lua
AmaranthineCodices: Oops, you're right!

LPGhatguy:
Going in!

PR #157, opened 2018-10-01
Roll back changes that allowed setState to be called inside willUpdate

### LINKED ISSUE #153: Roll back `willUpdate` part of #139
#139 loosened the restrictions of `setState` in both `willUpdate` and `init`. Doing the latter was a good change as it reduced confusion and disparity between assigning to `state` directly but only ever in `init` and calling `setState` everywhere else.

Making `willUpdate` allow `setState` was a grave mistake I think, however. I ran into this when working on the `new-reconciler` branch and reimplementing the `setState` variants. Should the new (not yet committed) state be used for state merges? Should a new update happen since `shouldUpdate` is getting skipped?

The case that came up internally that prompted me to consider this change would've been better solved by `getDerivedStateFromProps`. It should be adjusted to match current behavior in React 16.3, which would've addressed the same issue in a more elegant way.

ZoteTheMighty:
Closes #153.

Rolls back the changes to `Component` that allowed `setState` to be called during `willUpdate` . This helps avoid some edge cases in state updates where correct behavior was difficult to determine.

Checklist before submitting:

### DIFF
--- a/lib/Component.spec.lua
+++ b/lib/Component.spec.lua
@@ -434,19 +434,17 @@ return function()
 			end).to.throw()
 		end)
 
-		it("should only render once when called in willUpdate", function()
+		it("should throw when called in willUpdate", function()
 			local TestComponent = Component:extend("TestComponent")
-			local forceUpdate
 
+			local forceUpdate
 			function TestComponent:init()
 				forceUpdate = function()
 					self:_forceUpdate()
 				end
 			end
 
-			local renderCount = 0
 			function TestComponent:render()
-				renderCount = renderCount + 1
 				return nil
 			end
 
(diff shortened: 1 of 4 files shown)


lib/Component.spec.lua
```diff
@@ -456,17 +454,12 @@ return function()
...
-			forceUpdate()
-
-			expect(renderCount).to.equal(2)
+			local element = createElement(TestComponent)
 
-			Reconciler.unmount(handle)
+			expect(function()
+				Reconciler.mount(element)
+				forceUpdate()
+			end).to.throw()
```
LPGhatguy: This test will also pass if we rename `_forceUpdate`, an internal API. Can we refactor this test to use `Reconciler.reconcile` or whatnot instead?

ZoteTheMighty: Good point, I'll make that change.

Issue #158, opened 2018-10-19
Bindings: The sequel to Refs


ZoteTheMighty:
# Unidirectional Bindings

One of the biggest dangers and pain points of Roact is using refs. Refs are powerful, but fairly primitive in how they work. We can address this with an API for unidirectional bindings.

Bindings will be a more ergonomic version of refs that similarly allow data to be bound directly to Roblox instances.

They could also be used to implement things like gamepad navigation, which is awkward and difficult with the current state of refs.

### Roact.createBinding()

`createBinding(initialValue) -> Binding, updateFunction`

Bindings are typically created in a component's `init` method (the same way that `createRef` is used now).

The second return value of `createBinding` is a function that takes a value with which to update the binding. This function can be called to update a binding, which will update any values that are associated with it. When bindings are used as refs, this will only be needed by Roact itself.

### Binding methods

`binding:getValue() -> value`

Bindings provide a means of accessing their current value (equivalent to accessing `ref.current`).

`binding:map(mappingFunction)`

Creates a new binding by mapping the value of the existing one to a new value. This makes Bindings composable and is useful for transforming a value before assigning it to a Roblox Instance property.

## Refs

Refs can be re-implemented in terms of bindings. This is an internal detail, but it should demonstrate how bindings are nothing more than a modified version of refs.

PR #161, opened 2018-11-02
First-Class Refs for current reconciler

ZoteTheMighty:
## First Class Refs
Gamepad navigation with Roact is extremely painful, and one of the changes slated for the upcoming redesign (on the [new-reconciler](https://github.com/Roblox/roact/tree/new-reconciler) branch) is #158. This feature may be helpful by itself to alleviate some of these problems for the time being.

### Functionality
These changes allow refs to be provided as properties to Roact Host (or "Primitive") Components as proxies for other Roblox `Instance`s. This is particularly useful for properties like `GuiObject.NextSelectionLeft`, which expects another Roblox `Instance` as it's value. The reconciler will also create a subscription to the ref in question, so that changes to the ref update the property that subscribes to it.

### Implementation
This PR includes a minimal implementation of bindings that we'll refer to as First-Class Refs. It allows us to capture the lowest-hanging use case for bindings with minimal impact to Roact's current behavior.

For First-Class Refs, bindings are no more than an implementation detail of refs.

### Testing
Current testing on this branch is alright, but should be more thorough to determine that we don't modify behavior of any current valid use cases of `Roact.Ref` and `Roact.createRef`.

In particular, `Reconciler.lua` coverage probably needs to go up before we can be confident in this change.

### Checklist before submitting
* [x] More thorough testing (see above)
* [x] Added/updated documentation (add details to refs section)


lib/createRef.lua
```diff
@@ -1,20 +1,38 @@
...
+			if key == "current" then
+				return binding:getValue()
+			else
+				return binding[key]
+			end
+		end,
+		__newindex = function(self, key, value)
+			if key == "current" then
+				-- TODO: Confirm that this will not break any existing behavior
+				error("Assigning to ref.current is not allowed!", 2)
```
ZoteTheMighty: This needs further consideration. There may be places elsewhere in our code where we assign to `current`.

ZoteTheMighty:
This implementation seemed a little bit buggy and the need for this feature in the current master has probably passed, since we're pushing hard for `new-reconciler` to be merged to master soon.

PR #162, opened 2018-11-05
WIP: Scheduler interface for new reconciler

LPGhatguy:
This lays the foundations for having asynchronous rendering on top of the new reconciler.

The premise of this approach is simple:
- Mimic the same interface as the reconciler, but using less temporally certain terms like "schedule"
- Make user-triggered updates go through the scheduler instead of directly to the reconciler


LPGhatguy:
We should come back to this idea a bit later once we have a better understanding of this problem.

PR #163, opened 2018-11-14
WIP: Events for new reconciler

AmaranthineCodices:
Events for the new reconciler :tada:

Host virtual nodes will start handling events whenever an event is first used in the virtual node. If your node is static and has no events in its lifecycle, the event machinery is never set up. Events won't be fired at all during the initial mount process; any events fired in the middle of subsequent updates are queued and fired after the update is completed.

Checklist before submitting:
* [ ] More comments for explanation/internal documentation?
* Documentation updates aren't necessary, since this feature's already documented!


LPGhatguy:
This looks really good! I'm going to merge and play with it.

PR #165, opened 2018-11-20
New Reconciler: Fix possible error when processing boolean element

ZoteTheMighty:
Discovered this issue in the new reconciler when pointing it at a large existing project.

When we return a boolean as an element, mounting it will result in a nil virtual node. This is by design, but the `updateVirtualNodeChildren` function fails to account for it. Maybe there's a better place to account for this, or it can be resolved by updating depth information elsewhere?

Additionally, the `replaceVirtualNode` case doesn't appear to be technically possible, because it should always short-circuit beforehand.

~~I haven't yet written a regression test for this, but it should boil down to rendering an element tree in which a node below the root resolves to a boolean.~~

This is a pain to do with the NoopRenderer, and it feels weird to use the RobloxRenderer just to test the regression here, especially since the broken code has never been live anyway.

Checklist before submitting:
* ~~Added/updated relevant tests~~

### DIFF
--- a/lib/createReconciler.lua
+++ b/lib/createReconciler.lua
@@ -35,7 +35,11 @@ local function createReconciler(renderer)
 
 		unmountVirtualNode(virtualNode)
 		local newNode = mountVirtualNode(newElement, hostParent, hostKey)
-		newNode.depth = depth
+
+		-- mountVirtualNode can return nil if the element is a boolean
+		if newNode ~= nil then
+			newNode.depth = depth
+		end
 
 		return newNode
 	end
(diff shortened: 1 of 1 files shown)


ZoteTheMighty:
This is a pain to test with the `NoopRenderer` because the concept of children only really makes sense when using host components or fragments. If we're okay with not testing it, this should be good to go.

LPGhatguy:
Maybe we should make another renderer that does slightly more than `NoopRenderer` (like have the concept of children in host nodes?) but still doesn't cause any side effects.

PR #166, opened 2018-11-20
New Reconciler: Fix case where context is not passed.

ZoteTheMighty:
Discovered this issue in the new reconciler when pointing it at a large existing project.

Context was not being carried over during an update if the component rendered with a given key was changed to a new component type (a case in which we currently warn, but expect everything to work fine).

Writing a regression test for this should be pretty straightforward, but requires several steps.

Checklist before submitting:

### DIFF
--- a/lib/Component.spec/context.spec.lua
+++ b/lib/Component.spec/context.spec.lua
@@ -140,10 +140,70 @@ return function()
 		assertDeepEqual(capturedContext, expectedContext)
 	end)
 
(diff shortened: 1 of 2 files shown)


ZoteTheMighty:
Oh hang on, this won't work as it's written now. It would end up carrying over context created by the node that was replaced (instead of inheriting the parent's context).

I'll figure out a better solution...

LPGhatguy:
I like this solution!

PR #167, opened 2018-11-21
New Reconciler: boolean event fix

ZoteTheMighty:
Fixes slight functionality divergence; the old reconciler let users set [Roact.Event.Whatever] properties to a boolean, which would behave the same way as disconnecting or not connecting them (similar to boolean elements).

Note that the old code does this outside of the event manager, in the reconciler itself. Because of the way the connections were refactored to connect/connect-with-nil instead of connect/disconnect, it seemed to make more sense for the event manager to be the one to support it.

Checklist before submitting:


ZoteTheMighty:
Opting to fix the behavior in the old reconciler rather than carrying it over into the new one.

PR #172, opened 2018-12-03
WIP: Fragments for new reconciler

### LINKED ISSUE #7: Implement Fragment API

ZoteTheMighty:
Closes #7 (for the new reconciler).

I started from the `new-reconciler-fragments` branch that @LPGhatguy made and finished repairing/adding tests and patching inconsistencies.

One thing that fell out of this change was the need to distinguish between these two scenarios:
1. Updating a virtual node's children from a list of new children
   1. This is done by the `RobloxRenderer` to process the children of host nodes, and also the reconciler itself to process children specified via `Roact.Children`.
   2. The children are expected to be in the form of a table.
2. Updating a virtual node's children with the result of rendering a component.
   1. This is done by `Component` with its lifecycle methods, and also by the reconciler itself to mount/update function components.
   2. The expected format is either an element or a fragment (or a boolean value or nil).

Right now, I'm enforcing the above rules with two different reconciler functions for updating a node's children.

Checklist before submitting:

### DIFF
--- a/lib/ChildUtils.lua
+++ b/lib/ChildUtils.lua
@@ -95,7 +95,13 @@ function ChildUtils.getChildByKey(elements, hostKey)
 		return nil
 	end
 
-	return elements[hostKey]
+	if Type.of(elements) == Type.Fragment then
+		return elements.elements[hostKey]
+	end
+
+	error("Invalid elements")
+	-- TODO: Let's fix this case; it should work for fragments(?), but not arbitrary tables
+	-- return elements[hostKey]
 end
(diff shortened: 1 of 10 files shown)


LPGhatguy:
Fragments are a go!

PR #173, opened 2018-12-18
Error boundaries!

AmaranthineCodices:
Adds error boundaries like React's newer version of them. An error boundary is any component that declares a method with the signature `static getDerivedStateFromError(message: string) -> setStateMapper`. If this method is declared, the component will capture any errors from its children, pass the message to `getDerivedStateFromError`, apply the resulting state delta, then re-try the render.

The boundary will re-try the render only once; if the re-tried render throws, that error will propagate upwards until it either surfaces or hits another error boundary. It is the responsibility of the error boundary component to make sure this doesn't happen. If the re-render succeeds, the normal lifecycle process (calling either `didMount` or `didUpdate` depending on which is appropriate) will continue.

Questions I'm not sure about:
* Is this actually safe, or does this secretly corrupt some state that causes problems later?
* Should `getDerivedStateFromError` be allowed to return a mapper function, or should it only be allowed to return a dictionary? Allowing a function to be returned gives the function, indirectly, access to the component's state and props tables.

Checklist before submitting:


lib/Component.spec/getDerivedStateFromError.spec.lua
```diff
@@ -0,0 +1,190 @@
...
+        local key = "Test"
+
+        -- This will throw, because we don't stop rendering the throwing
+        -- component in response to the error. We test this more
+        -- rigorously elsewhere; this is just to keep the test from failing.
+        expect(function()
+            noopReconciler.mountVirtualNode(element, hostParent, key)
+        end).to.throw()
+        expect(callCount).to.equal(1)
+    end)
```
ZoteTheMighty: You should use `createSpy` for these tests! I think it might reduce a lot of your boilerplate, especially if you use `spy.assertCalledWith`

lib/Component.lua
```diff
@@ -121,6 +121,44 @@ function Component:render()
 	error(message, 0)
 end
 
+--[[
+	Tries to update the component's children, correctly handling the presence of
+	an error boundary.
+]]
+function Component:__updateChildren(reconciler, virtualNode)
```
ZoteTheMighty: Not 100% sure, but is this function name a little misleading? The function's job is basically "render and handle errors", and the "render" part of it is, I think, not quite the same thing as updating children?

I don't really have any suggestions, though... just wanted to see what you thought.

AmaranthineCodices: I picked it largely because when I initially wrote this it was an error handling wrapper around `reconciler.updateVirtualNodeChildren` - after I merged in the latest upstream changes that method went away, but the name for this didn't change. I've changed it to `__updateVirtualNode` to reflect it, but I don't have any super insightful reasoning for this name either!

lib/Component.spec/getDerivedStateFromError.spec.lua
AmaranthineCodices: Yeah! `assertCalledWith` isn't super useful here but I can use it in #174.

lib/Component.lua
```diff
@@ -186,6 +188,36 @@ function Component:render()
...
+			renderResult = instance:render()
+			internalData.lifecyclePhase = ComponentLifecyclePhase.ReconcileChildren
+			-- Don't try to handle errors at this point - the component should
+			-- be in a position to render a non-throwing fallback by now.
+			reconciler.updateVirtualNodeWithRenderResult(virtualNode, hostParent, renderResult)
+		end
+	else
+		reconciler.updateVirtualNodeWithRenderResult(virtualNode, hostParent, renderResult)
+	end
+end
```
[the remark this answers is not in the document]
AmaranthineCodices: Yeah, I need to revisit this when I have the time - there are definitely cases that error boundaries _should_ catch, but don't.

PR #174, opened 2018-12-18
validateProps for new reconciler

### LINKED ISSUE #24: Opt-in PropTypes implementation

AmaranthineCodices:
Supersedes #94; closes #24. Functionally the same as #94, but for the new reconciler instead! Briefly, this adds a new static method to components:

```
static validateProps(props: table) -> (false, message: string) | true
```

Every time props are updated, `validateProps` will be called with the new props before `shouldUpdate` and `init` are called. `validateProps` returns the same parameters that `assert` expects: a boolean, true if the props passed validation, false if they did not, plus a message explaining why they failed. If the first return value is true, the second value is ignored.

This is intended to be used with a type checking library like [PropTypes](https://github.com/AmaranthineCodices/rbx-prop-types) or [t](https://github.com/osyrisrblx/t).

One question remains about this: should this use GlobalConfig to conditionally disable it? I know we're looking to move away from GlobalConfig, so I'm not sure!

Checklist before submitting:

### DIFF
--- a/lib/Component.spec/validateProps.spec.lua
+++ b/lib/Component.spec/validateProps.spec.lua
@@ -1,0 +1,135 @@
+return function()
+    local createElement = require(script.Parent.Parent.createElement)
+    local createReconciler = require(script.Parent.Parent.createReconciler)
+    local createSpy = require(script.Parent.Parent.createSpy)
+    local NoopRenderer = require(script.Parent.Parent.NoopRenderer)
+
+    local Component = require(script.Parent.Parent.Component)
+
+    local noopReconciler = createReconciler(NoopRenderer)
+
+    it("should be invoked when mounted", function()
+        local MyComponent = Component:extend("MyComponent")
+
+        local validatePropsSpy = createSpy(function()
+            return true
+        end)
+
+        MyComponent.validateProps = validatePropsSpy.value
+
+        function MyComponent:render()
+            return nil
+        end
+
+        local element = createElement(MyComponent)
+		local hostParent = nil
+        local key = "Test"
+
+        noopReconciler.mountVirtualNode(element, hostParent, key)
+        expect(validatePropsSpy.callCount).to.equal(1)
+    end)
+
+    it("should be invoked when props change", function()
+        local MyComponent = Component:extend("MyComponent")
+
+        local validatePropsSpy = createSpy(function()
+            return true
+        end)
+
+        MyComponent.validateProps = validatePropsSpy.value
+
+        function MyComponent:render()
+            return nil
+        end
+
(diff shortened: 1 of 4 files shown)


AmaranthineCodices:
New concern: is there a better/shorter name than `assertCalledWithDeepEqual`?

lib/Component.spec/validateProps.spec.lua
```diff
@@ -0,0 +1,172 @@
...
+        end)
+
+        MyComponent.validateProps = validatePropsSpy.value
+
+        function MyComponent:render()
+            return nil
+        end
+
+        local element = createElement(MyComponent)
+		local hostParent = nil
```
ZoteTheMighty: This file has some weird mixed whitespace shenanigans going on. Let's clean that up!

lib/Component.spec/validateProps.spec.lua
```diff
@@ -0,0 +1,172 @@
...
+        local key = "Test"
```
LPGhatguy: Indented with spaces! Check your editor settings!

PR #175, opened 2019-01-02
New Reconciler: Document bindings

ZoteTheMighty:
Follows up #159 

Add documentation and changelog entry for bindings feature. This is a pretty quick draft, so it may currently be unclear or use some inconsistent wording. Most of the "Refs" section of the docs have been left alone, but I made some minor updates to acknowledge their relationship to bindings.

Checklist before submitting:
* ~Added/updated relevant tests~


docs/api-reference.md
```diff
@@ -59,6 +59,33 @@ If `children` contains more than one child, `oneChild` function will throw an er
...
+The first value returned is a `Binding` object, which will typically be passed as a prop to a Roact host component. The second is a function that can be called with a new value to update the binding.
+
+A `Binding` has the following API:
+
+#### getValue
+```
+Binding:getValue() -> value
+```
+
+Returns the internal value. This isn't typically necessary, but can be helpful when updating a binding relative to its current value.
```
LPGhatguy: We should note that using `getValue()` in render is probably a bad idea (for now) since your component won't re-render.

PR #176, opened 2019-01-03
New rec document fragments

ZoteTheMighty:
Follows up #172

Rough draft of some documentation for fragments. This example may not be particularly clear, and more examples or explanations are probably needed.

Checklist before submitting:
* ~Added/updated relevant tests~


ZoteTheMighty:
Hmm, not understanding what I'm seeing as the list of commits here. 60% of these commits don't make sense and should be already incorporated... I'll see if I can fix it.

PR #178, opened 2019-01-30
Update Config.lua

howmanysmall:
You didn't really need to localized that message, nor did you need to format. Just use concatenation.


LPGhatguy:
We're going to be getting rid of `GlobalConfig` before `new-reconciler` lands, so I don't think there's much reason to work on tidying it up right now.

PR #181, opened 2019-02-05
Remove type change warning

AmaranthineCodices:
Removes the warning when a component's type changes. We've found some very valid use cases for this, and the warning doesn't make much sense in light of them.


LPGhatguy:
very aggressive branch name, I like it

PR #183, opened 2019-02-15
setState suspension

### LINKED ISSUE #180: setState suspension to replace event suspension
This example causes a detailed error message in `master` and causes a stack overflow in `new-reconciler`:

```lua
local Child = Component:extend("Child")

function Child:render()
	return nil
end

function Child:didMount()
	self.props.callback()
end

local Parent = Component:extend("Parent")

function Parent:render()
	return createElement(Child, {
		callback = function()
			self:setState({})
		end,
	})
end
```

We should replace the event suspension mechanism we already put into the new reconciler with a more general `setState` deferring strategy that should look similar.

This blows up in a number of ways in real code that we've shipped, but this (hopefully magical) solution should be the end-all for the dreaded error message:

```
setState cannot be called while a component is being reified or reconciled.
This is the step where Roact constructs Roblox instances, and starting another
render here would introduce bugs.

Check the component %q to see if setState is being called by:
* a child Ref
* a child Changed event
* a child's render method
```

ZoteTheMighty:
Closes #180.

Work in progress for `setState` suspension. So far I've gotten a quick'n'dirty implementation in, and I've culled the similar functionality from `SingleEventManager`.

Wondering if a slightly better approach would just be to track *all* lifecycle states as one big enumerated set and then, in `setState`, granularly react based on it. (e.g. `if lifecyclePhase == shouldUpdate or lifecyclePhase == willUpdate or...`). Then when we are supposed to error, we can do so with the name known. Not that different from tracking `setStateBlockedReason` as we did previously.

Also, fixes a broken test from before (not sure how this happened) where `createFragment` wasn't added to the public api test.

Some basic tests are written, but more need to be written for cases like:
* [x] Ensure that multiple pending state updates are combined properly
* [x] Test interactions between suspension and `setState` aborting (via functional `setState`) functionality
* [x] Test interactions with `getDerivedStateFromProps`

Checklist before submitting:
* [x] Update error messages for disallowed rendering
* [x] Clean up or move no-longer-relevant tests for SingleEventManager
* [x] Change the LifecyclePhase object to be more enum-like
* [ ] Added/updated documentation (we should explain `setState` suspension somewhere in the docs)


ZoteTheMighty:
I'd still like to add more tests, but this is starting to get to a place where I'd like some eyes on it and some feedback.

The challenging bits were with `getDerivedStateFromProps` and making sure to avoid creating new state tables until I was sure state had actually changed. Maybe there are still better ways to simplify these things.

Also, better method name suggestions are always welcome.

PR #184, opened 2019-03-01
Tagging

AmaranthineCodices:
Lets you tag rendered Roact instances with `Roact.Tag`:

```lua
local function Test(props)
        BackgroundColor3 = props.backgroundColor,
        [Roact.Tag] = "Test"
    })
end
```

**Currently needs lemur changes for unit tests to pass.**


LPGhatguy:
@tiffany352 mentioned that it would be useful to assign multiple tags to the same instance.

We could accomplish that with a key like `Roact.Event`:

```lua
Roact.createElement("Frame", {
    [Roact.Tag.Foo] = true,
    [Roact.Tag.Bar] = true,
})
```

We would choose this over a list of tags because:
- Diffing lists quickly is difficult
- Creating an extra table per render is heavy
- Accepting a string or list makes the API more confusing/complicated to implement

PR #193, opened 2019-04-26
Update terminology in documentation

ZoteTheMighty:
* Remove references to `primitive` components/elements and use `host` as the descriptor
* Similarly, we should change `functional component` -> `function component`
* Update some links that appear to be broken???

Checklist before submitting:
* ~~Added/updated relevant tests~~

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -226,8 +226,8 @@ It's similar to `Roact.Event`:
 
 ```lua
 Roact.createElement("ScrollingFrame", {
-	[Roact.Change.CanvasPosition] = function(rbx, position)
-		print("ScrollingFrame scrolled to", position)
+	[Roact.Change.CanvasPosition] = function(rbx)
+		print("ScrollingFrame scrolled to", rbx.CanvasPosition)
 	end,
 })
 ```
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -111,7 +111,7 @@ Creates a new reference object that can be used with [Roact.Ref](#roactref).
 ### Roact.Children
 This is the key that Roact uses internally to store the children that are attached to a Roact element.
 
-If you're writing a new functional or stateful element that needs to be used like a host component, you can access `Roact.Children` in your props table.
+If you're writing a new function component or stateful component that renders children like a host component, you can access `Roact.Children` in your props table.
 
 ### Roact.Ref
 Use `Roact.Ref` as a key into the props of a host element to receive a handle to the underlying Roblox Instance.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -36,19 +36,19 @@ Creates a Roblox Instance given a Roact element, and optionally a `parent` to pu
 
 The result is a `ComponentInstanceHandle`, which is an opaque handle that represents this specific instance of the root component. You can pass this to APIs like `Roact.unmount` and the future debug API.
 
-### Roact.reconcile
+### Roact.update
 ```
-Roact.reconcile(instanceHandle, element) -> ComponentInstanceHandle
+Roact.update(instanceHandle, element) -> ComponentInstanceHandle
 ```
 
 Updates an existing instance handle with a new element, returning a new handle. This can be used to update a UI created with `Roact.mount` by passing in a new element with new props.
 
-`reconcile` can be used to change the props of a component instance created with `mount` and is useful for putting Roact content into non-Roact applications.
+`update` can be used to change the props of a component instance created with `mount` and is useful for putting Roact content into non-Roact applications.
 
 !!! warning
-	`Roact.reconcile` takes ownership of the `instanceHandle` passed into it and may unmount it and mount a new tree!
+	`Roact.update` takes ownership of the `instanceHandle` passed into it and may unmount it and mount a new tree!
 
-	Make sure to use the handle that `reconcile` returns in any operations after `reconcile`, including `unmount`.
+	Make sure to use the handle that `update` returns in any operations after `update`, including `unmount`.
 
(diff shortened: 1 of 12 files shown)


docs/performance/reduce-reconciliation.md
```diff
@@ -82,7 +82,7 @@ Now, if we add a new item to the end of the `Inventory` or change something abou
 
 Another performance improvement we can make is to use stable, unique keys to refer to our child elements.
 
-When the list that we pass into the `Inventory` component changes, Roact reconciles our Roblox UI by adjusting the properties of each primitive according to the new list of elements.
+When the list that we pass into the `Inventory` component changes, Roact reconciles our Roblox UI by adjusting the properties of each Roblox Instance according to the new list of elements.
```
LPGhatguy: Let's drop the verb 'reconcile' here and just say 'update'. We should probably do that across the board!

ZoteTheMighty: Any thoughts on the page name and some of our other references in here? `Roact.update` is better than `Roact.reconcile`, but I do still think that reconciliation is helpful as a term because it seems a little more specific.

Agreed that "update" is better here, though. I'll change it.

LPGhatguy: Not sure. Reconciliation is probably fine!

docs/api-reference.md
```diff
@@ -36,19 +37,19 @@ Creates a Roblox Instance given a Roact element, and optionally a `parent` to pu
 
 The result is a `ComponentInstanceHandle`, which is an opaque handle that represents this specific instance of the root component. You can pass this to APIs like `Roact.unmount` and the future debug API.
 
-### Roact.reconcile
+### Roact.update
 ```
-Roact.reconcile(instanceHandle, element) -> ComponentInstanceHandle
+Roact.update(instanceHandle, element) -> ComponentInstanceHandle
 ```
```
LPGhatguy: Let's add a note similarly to unmount/mount about the deprecated alias.

docs/api-reference.md
```diff
@@ -106,15 +108,49 @@ Roact.createRef() -> Ref
...
+```
+
+The entry point for configuring Roact. Roact currently applies this to everything using this instance of Roact, so be careful using this with a project that has multiple consumers of Roact.
+
+Once config values are set, they will apply from then on. This is primarily useful when developing as it can enable features that validate your code more strictly. Most of the settings here incur a performance cost and should typically be disabled in production environments.
+
+The following are the valid config keys that can be used, and what they do.
+
+#### typeChecks
+
```
LPGhatguy: We should prune these extra newlines to make it more clear where sections are when reading in plaintext.

docs/api-reference.md
```diff
@@ -149,15 +185,15 @@ Roact.createElement("Frame", {
...
 
 !!! warning
-	When `Roact.Ref` is given a function, Roact does not guarantee when this function will be run relative to the reconciliation of other props. If you try to read a Roblox property that's being set via a Roact prop, you won't know if you're reading it before or after Roact reconciles that prop!
+	When `Roact.Ref` is given a function, Roact does not guarantee when this function will be run relative to the reconciliation of other props. If you try to read a Roblox property that's being set via a Roact prop, you won't know if you're reading it before or after Roact updates that prop!
 
 !!! warning
 	When `Roact.Ref` is given a function, it will be called with `nil` when the component instance is destroyed!
 
-See [the refs guide](/advanced/refs.md) for more details.
+See [the refs guide]..(/advanced/refs) for more details.
```
LPGhatguy: Looks like a bad .. slipped in here!

docs/api-reference.md
```diff
@@ -363,6 +399,21 @@ By default, components are re-rendered any time a parent component updates, or w
...
+*Added in 0.2.0*
+```
+static validateProps(props) -> (false, message: string) | true
+```
+
+`validateProps` is an optional method that can be implemented for a component. It provides a mechanism for verifying inputs passed into the component.
+
+Every time props are updated, `validateProps` will be called with the new props before proceeding to `shouldUpdate` or `init`. It should return the same parameters that assert expects: a boolean, true if the props passed validation, false if they did not, plus a message explaining why they failed. If the first return value is true, the second value is ignored.
+
+**For performance reasons, property validation is disabled by default.** To use this feature, enable it via [`setGlobalConfig`](#roactsetglobalconfig).
```
LPGhatguy: Can we show a little code snippet of how to enable it?

docs/guide/state-and-lifecycle.md
```diff
@@ -39,11 +39,11 @@ Right now, this version of `setState` works exactly the same way as the version
 ## Lifecycle Methods
 Stateful components can provide methods to Roact that are called when certain things happen to a component instance.
 
-Lifecycle methods are a great place to send off network requests, measure UI ([with the help of refs](/advanced/refs)), wrap non-Roact components, and produce other side-effects.
+Lifecycle methods are a great place to send off network requests, measure UI ([with the help of refs](../../advanced/bindings-and-refs.md)), wrap non-Roact components, and produce other side-effects.
```
LPGhatguy: File extension was added here, intentional?

lib/Component.lua
```diff
@@ -177,6 +177,7 @@ end
 ]]
 function Component:getElementTraceback()
 	return self[InternalData].virtualNode.currentElement.source
+		or "<enable element tracebacks>"
```
LPGhatguy: I think we want this to return nil if source isn't present, otherwise it's not feasible for consumers of `getElementTraceback` to actually tell if tracebacks are enabled.

Beyond that, I think that's out of the scope of this PR.

PR #204, opened 2019-05-04
joinBindings implementation

Reselim:
Implements a function that creates a new binding, which updates whenever any of the bindings passed (via table) updates. Internally as `Binding.join`; public API is `Roact.joinBindings`.

Example usage:

```lua
local colorBinding = Roact.joinBindings({
	hue = hueBinding,
	saturation = saturationBinding,
	value = valueBinding,
}):map(function(values)
	return Color3.fromHSV(values.hue, values.saturation, values.value)
end)
```

### DIFF
--- a/src/Binding.lua
+++ b/src/Binding.lua
@@ -206,6 +206,17 @@ end
 	be exposed to users of Roact.
 ]]
 function Binding.join(bindings)
+	if config.typeChecks then
+		assert(typeof(bindings) == "table", "Bad arg #1 to Binding.join: expected table")
+
+		for key, binding in pairs(bindings) do
+			--[[
+				Bindings exclusively and always have a InternalData member
+			]]
+			assert(binding[InternalData], ("Non-binding value passed into Binding.join at index %q"):format(key))
+		end
+	end
+
 	local joinedBinding = Binding.create(mapBindingsToValues(bindings))
 	local internalData = joinedBinding[InternalData]
 
--- a/src/Binding.lua
+++ b/src/Binding.lua
@@ -108,9 +108,11 @@ end
 function bindingPrototype:__upstreamDisconnect()
 	local internalData = self[InternalData]
 
-	for _, disconnect in pairs(internalData.upstreamConnections) do
+	for _, disconnect in ipairs(internalData.upstreamConnections) do
 		disconnect()
 	end
+
+	internalData.upstreamConnections = {}
 end
 
 --[[
--- a/src/Binding.lua
+++ b/src/Binding.lua
@@ -12,6 +12,13 @@ local function identity(value)
 end
 
 --[[
+	Determines if a table (array or dictionary) is empty
+]]
+local function isTableEmpty(value)
+	return not next(value)
+end
+
+--[[
 	Maps a table of bindings to their respective values. Used in Binding.join.
 ]]
 local function mapBindingsToValues(bindings)
@@ -50,7 +57,7 @@ function bindingPrototype:getValue()
 		This allows us to avoid subscribing to our source until someone
 		has subscribed to us, and avoid creating dangling connections.
 	]]
-	if internalData.upstreamBindingCount > 0 then
+	if not isTableEmpty(internalData.upstreamBindings) then
 		return internalData.valueTransform(self:__getValueFromUpstreamBindings())
 	end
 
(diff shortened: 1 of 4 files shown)


src/Binding.lua
```diff
@@ -146,4 +159,33 @@ function Binding.create(initialValue)
...
+	Creates a new binding which updates when any of the upstream bindings
+	updates, which can be further mapped into any value. This function will
+	be exposed to users of Roact.
+]]
+function Binding.join(bindings)
+	local joinedBinding, setter = Binding.create(mapBindingsToValues(bindings))
+	local internalData = joinedBinding[InternalData]
+
+	local upstreamConnections = {}
+	internalData.upstreamConnections = upstreamConnections
```
LPGhatguy: Can we change `upstreamBinding` to work through this same mechanism (a table instead of a single entry) to consolidate the disconnection logic?

Some things that would fall out of this:
1. `upstreamDisconnect` can become static instead of being a closure we generate in a couple places
2. `Binding:map` changes to fill `upstreamConnections` instead of setting `upstreamBinding`

src/Binding.spec.lua
```diff
@@ -21,6 +21,47 @@ return function()
 		end)
 	end)
 
+	describe("Binding.join", function()
```
LPGhatguy: Don't forget to modify and test `getValue` with `join`!

Reselim:
A bug came to my attention while working on refactoring as per lpg's review. `Binding.join` hooks up connections before it's subscribed to for the first time; this will be fixed in an upcoming commit.

Reselim:
Everything should be working well now! Excuse any stylistic issues as I'm tired, however there shouldn't be many (or any at all).

src/Binding.lua
```diff
@@ -132,8 +184,10 @@ function Binding.create(initialValue)
 			subscriberCount = 0,
 
 			valueTransform = identity,
-			upstreamBinding = nil,
-			upstreamDisconnect = nil,
+			isJoinedBinding = false,
+			upstreamBindings = {},
+			upstreamConnections = {},
+			upstreamBindingCount = 0,
```
AmaranthineCodices: We only ever use this value to check if the binding has _any_ upstream bindings; we never use the actual count. Could this be a boolean instead?

src/Binding.lua
```diff
@@ -53,13 +66,49 @@ function bindingPrototype:map(valueTransform)
...
+	return newValue
+end
+
+--[[
+	Disconnects all connections to upstream bindings
+]]
+function bindingPrototype:__upstreamDisconnect()
+	local internalData = self[InternalData]
+
+	for _, disconnect in pairs(internalData.upstreamConnections) do
```
AmaranthineCodices: We should clear out `upstreamConnections` in this method, either by emptying the array as we iterate through it or just by replacing it with a new array. I'm not sure which one is preferable in this case, or if it really matters.

src/Binding.lua
Reselim: Now checking if the `upstreamBindings` table is empty or not.

ZoteTheMighty:
One thing I forgot about is changelog/documentation (as @AmaranthineCodices mentioned). We'll want an API reference entry at least.

That brings up the question of versions, though, so we probably need to figure out how we want to operate going forward, if we want to have something like a development branch. We probably shouldn't add new functionality to master without incrementing the version.

@lpghatguy what do you think we should do about this for now?

src/Binding.lua
```diff
@@ -146,4 +205,33 @@ function Binding.create(initialValue)
...
+]]
+function Binding.join(bindings)
+	if config.typeChecks then
+		assert(typeof(bindings) == "table", "Bad arg #1 to Binding.join: expected table")
+
+		for key, binding in pairs(bindings) do
+			--[[
+				Bindings exclusively and always have a InternalData member
+			]]
+			assert(binding[InternalData], ("Non-binding value passed into Binding.join at index %q"):format(key))
```
AmaranthineCodices: We use `InternalData` in several data structures. You should use `Type.of` for this kind of type check:
https://github.com/Roblox/roact/blob/41af74df730723ed39c02f54cae7c6c97c6c5d9c/src/createReconciler.lua#L248
```suggestion
			assert(Type.of(binding) == Type.Binding, ("Non-binding value passed into Binding.join at index %q"):format(key))
```

Reselim: I was under the assumption that a new `Symbol` was created every time. Will fix.

LPGhatguy:
@ZoteTheMighty I think we'll probably operate like other projects do -- keep `master` as our dev branch and only increment the version number when we do releases.

I'd like to mess with this PR a little bit to see if it makes sense to split our different kinds of bindings apart now, and just keep a common interface.

LPGhatguy:
After talking on Discord and doing some hacking, I think we're going to close this in favor of #208.

PR #208, opened 2019-05-16
Refactor bindings, introducing joinBindings

LPGhatguy:
Inspired by #204. 😄 

Reworks bindings from the ground up and adds support for joining bindings together.

Checklist before submitting:

### DIFF
--- a/src/Binding.lua
+++ b/src/Binding.lua
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
 
 local config = require(script.Parent.GlobalConfig).get()
 
---[[
-	Default mapping function used for non-mapped bindings
-]]
-local function identity(value)
-	return value
+local BindingImpl = Symbol.named("BindingImpl")
+
+local BindingInternalApi = {}
+
+local BindingPublicMethods = {}
+
+function BindingPublicMethods:getValue()
+	return BindingInternalApi.getValue(self)
 end
 
-local Binding = {}
-
---[[
-	Set of keys for fields that are internal to Bindings
-]]
-local InternalData = Symbol.named("InternalData")
-
-local bindingPrototype = {}
-bindingPrototype.__index = bindingPrototype
-bindingPrototype.__tostring = function(self)
-	return ("RoactBinding(%s)"):format(tostring(self[InternalData].value))
+function BindingPublicMethods:map(predicate)
+	return BindingInternalApi.map(self, predicate)
 end
 
---[[
-	Get the current value from a binding
-]]
-function bindingPrototype:getValue()
-	local internalData = self[InternalData]
+local BindingPublicMeta = {
+	__index = BindingPublicMethods,
+	__tostring = function(self)
+		return string.format("RoactBinding(%s)", tostring(self:getValue()))
+	end,
+}
 
-	--[[
-		If our source is another binding but we're not subscribed, we'll
-		return the mapped value from our upstream binding.
+function BindingInternalApi.update(binding, newValue)
+	return binding[BindingImpl].update(newValue)
+end
 
-		This allows us to avoid subscribing to our source until someone
-		has subscribed to us, and avoid creating dangling connections.
-	]]
-	if internalData.upstreamBinding ~= nil and internalData.upstreamDisconnect == nil then
-		return internalData.valueTransform(internalData.upstreamBinding:getValue())
+function BindingInternalApi.subscribe(binding, callback)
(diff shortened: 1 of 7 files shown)


src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
+	function impl.subscribe(callback)
+		return BindingInternalApi.subscribe(upstreamBinding, function(newValue)
+			callback(predicate(newValue))
 		end)
 	end
 
-	local disconnect = internalData.changeSignal:subscribe(handler)
-	internalData.subscriberCount = internalData.subscriberCount + 1
+	function impl.update(newValue)
+		-- This operation doesn't make sense; we can leave it as a no-op
```
AmaranthineCodices: Let's throw here to catch weird cases (like using a mapped binding as a ref)

src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
+end
+
+function BindingInternalApi.join(upstreamBindings)
+	if config.typeChecks then
+		assert(typeof(upstreamBindings) == "table", "Expected arg #1 to be of type table")
+
+		for key, value in pairs(upstreamBindings) do
+			if Type.of(value) ~= Type.Binding then
+				local message = (
+					"Expected arg #1 to contain only bindings, but key %s had a non-binding value"
```
AmaranthineCodices: Should we use `%q` here, in case the key has spaces or something similarly odd?

src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
 
-	local setter = function(newValue)
-		Binding.update(binding, newValue)
+			disconnects = nil
+		end
 	end
 
-	return binding, setter
+	function impl.update(newValue)
+		-- This operation doesn't make sense; we can leave it as a no-op
```
AmaranthineCodices: Should throw here too.

src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
---[[
-	Default mapping function used for non-mapped bindings
-]]
-local function identity(value)
-	return value
+local BindingImpl = Symbol.named("BindingImpl")
+
+local BindingInternalApi = {}
+
+local BindingPublicMethods = {}
```
Reselim: Should this be changed to `bindingPrototype` to be congruent with other classes?

src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
-end
 
---[[
-	Create a new binding object with the given starting value. This
-	function will be exposed to users of Roact.
-]]
-function Binding.create(initialValue)
-	local binding = {
-		[Type] = Type.Binding,
+	function impl.subscribe(callback)
```
Reselim: Would it make more sense to only subscribe to `upstreamBindings` once and create/destroy the upstream connections depending on if there are subscribers or not? The prior implementation does this.

Upstream connections could easily add up, considering bindings mapped from this one will invoke `subscribe` when subscribed to. A good example of this potentially becoming a problem would be when animating transparency.

src/Binding.lua
```diff
@@ -4,146 +4,156 @@ local Type = require(script.Parent.Type)
...
-
-		--[[
-			If our subscribers count drops to 0, we can safely unsubscribe from
-			our source binding
-		]]
-		if internalData.subscriberCount == 0 and internalData.upstreamDisconnect ~= nil then
-			internalData.upstreamDisconnect()
-			internalData.upstreamDisconnect = nil
+	local impl = {
+		disconnectMethods = nil,
```
Reselim: This variable is unused.

src/Binding.lua
LPGhatguy: Just got back into working on this!

I don't think it's a big deal to pass subscriptions up through to their parents. At the end of the day, someone is going to have to track the connection, and consolidating that at the top is the simplest approach!

LPGhatguy:
PR feedback was addressed and I added a changelog entry and (bad) documentation.

I'm not sure what a small sample for how to use `joinBindings` would be; the current one is quite large.

Reselim:
LGTM! Works well in my (rather large) UI codebase making use of `joinBindings`.

LPGhatguy:
Will merge this first thing Monday to cut 1.1!

PR #210, opened 2019-05-23
Remove children when updating a host node to an element with nil children

### LINKED ISSUE #209: Updating from having children to having `nil` children doesn't remove children
This is something I hit in Rojo trying to upgrade to Roact 1.0!

When an update causes a change from:

```lua
Roact.createElement("Parent", nil, { Child = Roact.createElement("Folder") })
```
to
```lua
Roact.createElement("Parent", nil, nil)
```

the `Child` folder won't be removed. If children is specified as an empty table instead, the object appears to be removed.

Here's the relevant component in Rojo: https://github.com/LPGhatguy/rojo/blob/1e130971269367da2f58c8a78b464cfcd26650af/plugin/src/Components/App.lua#L82-L142

Changing line 83 from `local children` to `local children = {}` fixes the Rojo bug.

AmaranthineCodices:
Closes #209. In ea822f62ec11be3f979344cc34e03b81e48b0b53 we introduced an optimization in RobloxRenderer that avoids calling into the reconciler to update children if the element's children are `nil`. This introduced #209, where when updating a tree with an element that has `nil` children (not an empty table), the existing children are not unmounted.

Checklist before submitting:
* [x] ~~Added/updated documentation~~

### DIFF
--- a/src/RobloxRenderer.lua
+++ b/src/RobloxRenderer.lua
@@ -271,7 +271,7 @@ function RobloxRenderer.updateHostNode(reconciler, virtualNode, newElement)
 	local children = newElement.props[Children]
 
 	if children ~= nil or oldProps[Children] ~= nil then
-		reconciler.updateVirtualNodeWithChildren(virtualNode, virtualNode.hostObject, newElement.props[Children])
+		reconciler.updateVirtualNodeWithChildren(virtualNode, virtualNode.hostObject, children)
 	end
 
 	if virtualNode.eventManager ~= nil then
--- a/src/RobloxRenderer.lua
+++ b/src/RobloxRenderer.lua
@@ -269,7 +269,6 @@ function RobloxRenderer.updateHostNode(reconciler, virtualNode, newElement)
 	end
 
 	local children = newElement.props[Children]
-
 	if children ~= nil or oldProps[Children] ~= nil then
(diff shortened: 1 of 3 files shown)


LPGhatguy:
This change looks good to me, but I want to wait for @ZoteTheMighty.

Can we pull benchmark numbers out for this change, too?

ZoteTheMighty:
This looks okay to me, but I have a couple comments to consider:

I did some quick benchmarking, and there's a somewhat consistent 20-25% slowdown on benchmarks for specifically mount and update operations. Note that mounting and updating big nested trees sees a much smaller effect. The performance effect is specifically on the leaves of these element trees.

That said, there are a couple things we might be able to do regain some of this performance and still fix the bug.
1. I'm not sure the mount case is actually going to cause this bug. We may be able to revert that bit and still pass our new test.
2. We might be able to do a more intelligent short circuit with the update case along the lines of `if children ~= nil or oldProps[Roact.Children] ~= nil then`. Might be worth experimenting with.

src/RobloxRenderer.lua
```diff
@@ -269,7 +269,8 @@ function RobloxRenderer.updateHostNode(reconciler, virtualNode, newElement)
 	end
 
 	local children = newElement.props[Children]
-	if children ~= nil then
+
```
LPGhatguy:
```suggestion
```

PR #214, opened 2019-06-07
Make Fragment an Element

### LINKED ISSUE #205: Fragments don't work when used inline
Repro:
```lua
local Roact = require(script.Parent.Parent.Parent.Roact)

return function(target)
    local element = Roact.createElement("Frame", {
        Size = UDim2.new(1, 0, 1, 0),
    }, {
        Contents = Roact.createFragment({}),
    })

    local handle = Roact.mount(element, target)

    return function()
        Roact.unmount(handle)
    end
end
```

It looks like fragments only work when they're returned from a component currently. My usecase is using them inline. For example, in JS I can write:

```tsx
return (
  <div>
    <h1>header</h1>
    <>{articleList}</>
    <span>footer</span>
  </div>
)
```

and I want to do something similar with Roact.

```
16:17:40.314 - ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler:321: Unknown ElementKind "nil"
16:17:40.314 - Stack Begin
16:17:40.314 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler', Line 321 - upvalue mountVirtualNode
16:17:40.315 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler', Line 88 - upvalue updateChildren
16:17:40.315 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler', Line 100 - field updateVirtualNodeWithChildren
16:17:40.315 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.RobloxRenderer', Line 215 - field mountHostNode
16:17:40.316 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler', Line 313 - upvalue mountVirtualNode
16:17:40.316 - Script 'ServerStorage.TracingPlugin._Index.Roact-e3b370a8-41af74df.Packages.Roact.createReconciler', Line 351 - field mount
16:17:40.316 - Script 'ServerStorage.TracingPlugin.PluginWidgets.Storybook.Repro', Line 10
16:17:40.317 - Stack End
```

jeparlefrancais:
Closes #205

Fragments can now be used directly as children of a component:
```lua
local function component()
    return Roact.createElement("Frame", {}, {
        fragments = Roact.createFragment({
            A = Roact.createElement("Frame", {}, {})
        })
    })
end
```
Turning Fragments into Elements make them have their own virtual node, so they can be reconciled as any other nodes.

I'm not so sure about how to write good tests for that, let me know what's the best way (not sure where to assert exactly).

Checklist before submitting:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Roact Changelog
 
 ## Unreleased Changes
+* Fixed a bug where fragments could not be used as child of a component. ([#214](https://github.com/Roblox/roact/pull/214))
 
 ## [1.1.0](https://github.com/Roblox/roact/releases/tag/v1.1.0) (June 3rd, 2019)
 * Fixed an issue where updating a host element with children to an element with `nil` children caused the old children to not be unmounted. ([#210](https://github.com/Roblox/roact/pull/210))
--- a/src/RobloxRenderer.spec.lua
+++ b/src/RobloxRenderer.spec.lua
@@ -4,6 +4,7 @@ return function()
 	local Children = require(script.Parent.PropMarkers.Children)
 	local Component = require(script.Parent.Component)
 	local createElement = require(script.Parent.createElement)
+	local createFragment = require(script.Parent.createFragment)
 	local createReconciler = require(script.Parent.createReconciler)
 	local createRef = require(script.Parent.createRef)
 	local createSpy = require(script.Parent.createSpy)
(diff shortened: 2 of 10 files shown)


ZoteTheMighty:
For testing, I think it should be possible to use the `mountVirtualNode` method of the reconciler to get back a tree of virtual nodes that you can introspect to some degree? I'm not sure if you'll run into issues with data contained in private internal tables, but I don't think you will.

Look at some of the other tests in `createReconciler.spec.lua` for ideas, but I think the way to go is to get a tree of nodes and `expect` things like `Type.of(node.currentElement)` and `hostKey` and `children`, etc.

As for what cases to test, I'd just try combinations of places you can put fragments? An element with fragments in its children is a good start.

I'd also test things that will have name duplication and make sure they work out correctly, but those tests are more involved.

ZoteTheMighty:
Closing and re-opening to trigger CI checks.

src/RobloxRenderer.spec.lua
```diff
@@ -686,6 +687,100 @@ return function()
...
+			local childAComponentSpy = createSpy(function(props)
+				return createElement("StringValue", {
+					Value = "A",
+				})
+			end)
+			local childBComponentSpy = createSpy(function(props)
+				return createElement("StringValue", {
+					Value = "B",
+				})
+			end)
```
LPGhatguy: Do these needs to be spies?

jeparlefrancais: I don't think they need, I could have asserted that they were called, but instead I removed them and asserted that the values are correctly created

src/RobloxRenderer.spec.lua
```diff
@@ -741,6 +741,14 @@ return function()
 			expect(parentChildren[1].ClassName).to.equal("StringValue")
 			expect(parentChildren[2].ClassName).to.equal("StringValue")
 
+			if parentChildren[1].Value ~= "A" then
+				expect(parentChildren[1].Value).to.equal("B")
+				expect(parentChildren[2].Value).to.equal("A")
+			else
+				expect(parentChildren[1].Value).to.equal("A")
+				expect(parentChildren[2].Value).to.equal("B")
+			end
```
ZoteTheMighty: I wonder if there's a more readable way to do this? Maybe something like iterate through children and add them to a map of Value -> count, then make sure you have a table that looks like:
```
{
   A = 1,
   B = 1,
}
```
Either way, you probably want a comment indicating that you expect one child with Value "A" and one child with Value "B" but order doesn't matter.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Roact Changelog
 
 ## Unreleased Changes
+* Fixed a bug where fragments could not be used as child of a component. ([#214](https://github.com/Roblox/roact/pull/214))
```
ZoteTheMighty: How about:
```suggestion
* Fixed a bug where fragments could not be used as children of an element or another fragment. ([#214](https://github.com/Roblox/roact/pull/214))
```

src/RobloxRenderer.spec.lua
```diff
@@ -686,6 +687,109 @@ return function()
...
+			local node = reconciler.mountVirtualNode(createFragment({}), hostParent, "test")
+
+			expect(#hostParent:GetChildren()).to.equal(0)
+
+			reconciler.unmountVirtualNode(node)
+
+			expect(#hostParent:GetChildren()).to.equal(0)
+		end)
+	end)
+
```
LPGhatguy: Can we add tests for fragments with multiple children, as well as fragments with zero children?

I think our default for testing fragments should be at least two children, since a fragment with one child is unlikely to see much use in reality!

src/RobloxRenderer.spec.lua
```diff
@@ -686,6 +687,109 @@ return function()
...
+			local childAComponent = function(props)
+				return createElement("StringValue", {
+					Value = "A",
+				})
+			end
+			local childBComponent = function(props)
+				return createElement("StringValue", {
+					Value = "B",
+				})
+			end
```
LPGhatguy: Do we need these components at all?

If we do, we should define them using `local function ChildA() ... end` and such to line up with conventions for declaring function components.

src/createReconciler.spec.lua
```diff
@@ -286,4 +287,35 @@ return function()
...
+		end)
+
+		it("should mount an empty fragment", function()
+			local emptyFragment = createFragment({})
+			local node = noopReconciler.mountVirtualNode(emptyFragment, nil, "test")
+
+			expect(node).to.be.ok()
+			expect(next(node.children)).to.never.be.ok()
+		end)
+	end)
```
LPGhatguy: This would be a good spot to have a test for multiple children of the fragment!

PR #216, opened 2019-06-11
Better error message for invalid changed hook name

### LINKED ISSUE #206: Misleading error when reconciling changed hooks
Repro:
```lua
local Roact = require(script.Parent.Parent.Parent.Roact)

return function(target)
	local element = Roact.createElement("UIListLayout", {
		[Roact.Change.AbsoluteSize] = function(rbx)
			print(rbx.AbsoluteContentSize)
		end
	})

	local handle = Roact.mount(element, target)

	return function()
		Roact.unmount(handle)
	end
end
```
The error:
```
23:19:38.633 - Error applying props:
23:19:38.634 - 	AbsoluteSize is not a valid property name.
23:19:38.634 - In element:
23:19:38.634 - <enable element tracebacks>
23:19:38.634 - 
```
Ideally it would say that AbsoluteSize is being used as an event name, instead of as a property name. It'd also be helpful if it showed the ClassName of the instance the property/event is attempting to be applied to.

jeparlefrancais:
Closes #206 

It seemed like an easy improvement. I made the error message as explicit as I could.

One thing I'd like to be double checked on, is that the protected call won't catch another error and put that message instead. The case I see that could be misleading is if `self._instance` is nil, but is that reasonably  possible?

Checklist before submitting:


src/SingleEventManager.lua
```diff
@@ -53,7 +53,14 @@ function SingleEventManager:connectEvent(key, listener)
 end
 
 function SingleEventManager:connectPropertyChange(key, listener)
-	local event = self._instance:GetPropertyChangedSignal(key)
+	local success, event = pcall(function()
+		return self._instance:GetPropertyChangedSignal(key)
+	end)
+
+	if not success then
+		error(("Cannot get property '%s' changed signal from %s"):format(tostring(key), self._instance.ClassName), 0)
```
ZoteTheMighty: I think the grammar here might be a bit awkward? What do you think about:
```suggestion
		error(("Cannot subscribe to changed signal: %q is not a property of %s"):format(tostring(key), self._instance.ClassName), 0)
```

src/SingleEventManager.lua
```diff
@@ -53,7 +53,17 @@ function SingleEventManager:connectEvent(key, listener)
 end
 
 function SingleEventManager:connectPropertyChange(key, listener)
-	local event = self._instance:GetPropertyChangedSignal(key)
+	local success, event = pcall(function()
+		return self._instance:GetPropertyChangedSignal(key)
+	end)
+
+	if not success then
+		error(("Cannot subscribe to changed signal: %q is not a property of %s"):format(
```
LPGhatguy: This isn't necessarily what the actual error is, since it could be a permissions problem. :(

Do we want to do some sort of really nasty string matching to try to fix those problems, or are we okay with that edge case?

jeparlefrancais: For the permissions problem, do you mean that `GetPropertyChangedSignal` could throw if you ask for a specific property that is guarded by a certain level? Because the function itself isn't behind any level right? If it's true then I am less concerned about it because it's almost true that the property does not exist, in that permission context let's say.

One thing we could do is print out the original error, like appending `(original error thrown: %s)` to the message and pass the returned error message.

Or we could simply append to the message `(or lacking permission level)` too.

And I agree that the string matching on an error message might be too fragile (and also not elegant)

PR #220, opened 2019-06-20
ShopPage Component example

jeparlefrancais:
This example aims to integrate some of Roact concepts together in a more realistic component that could exists in a Roblox game.

I make this draft pull request so anyone can already review the code itself to see if there are any misleading stuff going on that could be confusing or considered bad practice.

### DIFF
--- a/docs/tutorials/shop-page.md
+++ b/docs/tutorials/shop-page.md
@@ -1,0 +1,31 @@
+# Shop page
+_Goal: Create a page to sell developer products in a game._
+
+[add an image of the result we are expecting]
+
+## Top Component
+Let's start by creating the top level component, that will be the page containing the items for sale in the game.
+
+```lua
+
+```
+
+### Viewing the component
+
+
+## Items Grid
+
+
+## Product Item
+[remind state and props]
+
+### Adding some animations
+
+
+## Integration
+
+### In a Roact project
+Here explain how the component can be used my another component or how it can be mounted on it's own (like main.client.lua does)
+
+### In a non-Roact project
+Explain how the page could be mounted in an existing UI instances tree. That opens the subject that it is possible to slowly port an existing UI to use Roact part by part.
--- a/examples/shop-page/Components/ProductItem.lua
+++ b/examples/shop-page/Components/ProductItem.lua
@@ -1,0 +1,69 @@
+local MarketplaceService = game:GetService("MarketplaceService")
+local Players = game:GetService("Players")
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+
+local Roact = require(ReplicatedStorage.Roact)
+
+local ProductItem = Roact.Component:extend("ProductItem")
+
+--[[
+	Props: {
+		image: the image displayed
+		price: how much robux the product cost
+		productId: the developer product id
+
+		-- Optional props
+		order: the index at which the item should be shown. If not
+		       provided, the price will be used as the index.
+	}
+]]
+
+function ProductItem:init()
+	self:setState({
+		onMouseEnter = function()
+			-- grow the icon a little
+		end,
+		onMouseLeave = function()
+			-- go back to original size
+		end,
+		onActivated = function()
+			local props = self.props
+
+			MarketplaceService:PromptProductPurchase(Players.LocalPlayer, props.productId)
+		end
+	})
+	self.padding, self.updatePadding = Roact.createBinding(0)
+end
+
+function ProductItem:render()
+	local props = self.props
+	local state = self.state
+	local padding = self.padding
+
+	return Roact.createElement("ImageButton", {
+		BackgroundTransparency = 1,
+		Image = '',
+		LayoutOrder = props.order or props.price,
+		[Roact.Event.MouseEnter] = state.onMouseEnter,
+		[Roact.Event.MouseLeave] = state.onMouseLeave,
+	}, {
+		Icon = Roact.createElement("ImageLabel", {
+			AnchorPoint = Vector2.new(0.5, 0.5),
+			BackgroundTransparency = 1,
+			Image = props.image,
+			Position = UDim2.new(0.5, 0, 0.5, 0),
+			Size = UDim2.new(1, padding, 1, padding),
+		}),
+		PriceLabel = Roact.createElement("TextLabel", {
+			AnchorPoint = Vector2.new(0.5, 1),
+			BackgroundTransparency = 1,
+			Text = ("R$ %d"):format(props.price),
+			TextColor3 = Color3.fromRGB(10, 200, 10),
+			TextScaled = true,
+			Position = UDim2.new(0.5, 0, 1, 0),
+			Size = UDim2.new(1, 0, 0.3, 0),
+		}),
+	})
+end
+
+return ProductItem
--- a/examples/shop-page/Components/ShopPage.lua
+++ b/examples/shop-page/Components/ShopPage.lua
@@ -1,0 +1,77 @@
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+local Players = game:GetService("Players")
+
+local Components = ReplicatedStorage:WaitForChild("Components")
+
+local Roact = require(ReplicatedStorage.Roact)
+
+local ProductItemList = require(Components:WaitForChild("ProductItemList"))
+
+local excludePropsForFrame = {
+	items = true,
+	itemsPerRow = true,
+	itemAspectRatio = true,
+	padding = true
+}
+
+local ShopPage = Roact.Component:extend("ShopPage")
+
+function ShopPage:init()
+	self:setState({
+		cellSize = 100,
+		canvasHeight = 100,
+	})
+
+	self.onAbsoluteSizeChanged = function(frame)
+		local props = self.props
+		local state = self.state
+		local padding = props.padding or 0
+		local itemsPerRow = props.itemsPerRow
+
+		local totalWidth = frame.AbsoluteSize.X
+		local cellWidth = (totalWidth - padding * (itemsPerRow + 1)) / itemsPerRow
+		local cellHeight = cellWidth * props.itemAspectRatio
+		local rows = math.ceil(#props.items / itemsPerRow)
+		local canvasHeight = rows * cellHeight + padding * (rows + 1)
+
+		self:setState({
+			cellSize = cellWidth,
+			canvasHeight = canvasHeight,
+		})
+	end
+end
+
+function ShopPage:render()
+	local props = self.props
+	local state = self.state
+	local padding = props.padding or 0
+	local cellSize = state.cellSize
+	local canvasHeight = state.canvasHeight
+
+	local frameProps = {}
+
+	for key, value in pairs(props) do
+		if not excludePropsForFrame[key] then
+			frameProps[key] = value
+		end
+	end
+
+	frameProps.CanvasSize = UDim2.new(0, 0, 0, canvasHeight)
+	frameProps.ClipsDescendants = true
+	frameProps[Roact.Change.AbsoluteSize] = self.onAbsoluteSizeChanged
+
+	return Roact.createElement("ScrollingFrame", frameProps, {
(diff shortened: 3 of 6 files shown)


examples/shop-page/default.project.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "Roact Test Place",
+	"tree": {
+		"$className": "DataModel",
+		"ReplicatedFirst": {
+			"$className": "ReplicatedFirst",
+			"main": {
+				"$path": "main.client.lua"
+			}
+		},
```
LPGhatguy: Should we just use `StarterPlayer.StarterPlayerScripts` here instead?

examples/shop-page/Components/ShopPage.lua
```diff
@@ -0,0 +1,77 @@
...
+	local cellSize = state.cellSize
+	local canvasHeight = state.canvasHeight
+
+	local frameProps = {}
+
+	for key, value in pairs(props) do
+		if not excludePropsForFrame[key] then
+			frameProps[key] = value
+		end
+	end
```
LPGhatguy: I think this is generally an anti-pattern. I like to explicitly scope all the props intended for the container to a sub-table, which is less error-prone!

examples/shop-page/Components/ProductItem.lua
```diff
@@ -0,0 +1,69 @@
...
+end
+
+function ProductItem:render()
+	local props = self.props
+	local state = self.state
+	local padding = self.padding
+
+	return Roact.createElement("ImageButton", {
+		BackgroundTransparency = 1,
+		Image = '',
```
LPGhatguy: Double quotes!

examples/shop-page/Components/ProductItem.lua
```diff
@@ -0,0 +1,69 @@
...
+		image: the image displayed
+		price: how much robux the product cost
+		productId: the developer product id
+
+		-- Optional props
+		order: the index at which the item should be shown. If not
+		       provided, the price will be used as the index.
+	}
+]]
+
```
LPGhatguy: We should use `validateProps` for this instead of writing a comment all on its lonesome!

examples/shop-page/Components/ProductItem.lua
```diff
@@ -57,13 +46,37 @@ function ProductItem:render()
...
+		assert(
+			type(props.productId) == "string" or type(props.productId) == "number",
+			("props.productId should be a string or a number (got %q)"):format(type(props.productId))
+		)
+		assert(
+			props.order == nil or type(props.order) == "number",
+			("props.order should be a number (got %q)"):format(type(props.order))
+		)
+	end)
+end
```
LPGhatguy: We should move this to the top of the file!

Now that I'm looking at it in writing (without having `t`), I'm way less enthusiastic about typechecking here. :(

The motivation of my original comment about having types in a comment is that it's generally a poor practice that has caused rot in code internally. It's easy for those comments to get out of sync with the real expectations of the component, but much harder for `validateProps` to!

jeparlefrancais: Yes I agree that comments get easily out of sync, and I did not know about `validateProps` so thanks for getting me up to speed on that!

So you would not go with type checking in validate props? What would you propose in that case? 

I could add a value constraint on the price to be positive.

LPGhatguy: I'm not sure. I kinda wish we could just depend on `t` here.

examples/shop-page/Components/ShopPage.lua
```diff
@@ -0,0 +1,74 @@
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+
+local Components = ReplicatedStorage:WaitForChild("Components")
```
MisterUncloaked: do we need to call WaitForChild here since the component lives inside Components?

examples/shop-page/main.client.lua
```diff
@@ -0,0 +1,63 @@
...
+
+local ShopPage = require(Components:WaitForChild("ShopPage"))
+
+local screenGui = Instance.new("ScreenGui")
+screenGui.Parent = Players.LocalPlayer:WaitForChild("PlayerGui")
+
+local shopPage = Roact.createElement(ShopPage, {
+	items = {
+		{
+			identifier = "Red Visor",
```
MisterUncloaked: should we put this static data in a separate file? it may help readers focus on the mechanics of how to mount the client UI in this file

examples/shop-page/default.project.json
```diff
@@ -0,0 +1,21 @@
+{
+	"name": "Roact Test Place",
```
MisterUncloaked: should we commit the default.project.json? Folks looking at this example may not know what Rojo is. Alternatively we could use add a README to this folder as a plug for Rojo for folks who want to quickly sync this into their game.

examples/shop-page/main.client.lua
jeparlefrancais: Would you put only the items data in a separate file? Or also all the other props?

MisterUncloaked: just the items

examples/shop-page/default.project.json
jeparlefrancais: I think I'm going to remove that! I just removed the client script to instead make the example loadable with the others from the `place.project.json` rojo configuration.

examples/shop-page/Components/ShopPage.lua
jeparlefrancais: Nope, my bad!

docs/tutorials/shop-page.md
```diff
@@ -0,0 +1,396 @@
...
+end
+```
+
+## Adding Items
+
+In the previous section we created a scrolling frame that contains a UIGridLayout object. Now, we want to add elements that represent each item we sell in our shop. In order to be able to add a variable amount of items in our component, we will iterate on the list of products to create an element for each of them.
+
+To keep our component simple, let's divide this feature into different components:
+ - **ProductItem:** a card that shows the price of an item with its image and when clicked prompt the player to buy the product
+ - **ProductItemList:** iterate on the list of items to create each individual ProductItem element
```
vocksel: I think you could name these `Item` and `ItemList`. `ProductItem` sounds like you're saying the same thing twice and is a little clunky to read

docs/tutorials/shop-page.md
```diff
@@ -0,0 +1,396 @@
...
+	})
+end
+```
+
+Don't forget to clean up! When the component will be unmounted, we need to destroy the Tween objects that were created. We use the [`willUnmount`](https://roblox.github.io/roact/guide/state-and-lifecycle/#lifecycle-methods) lifecycle method to destroy the objects.
+
+```lua
+function ProductItem:willUnmount()
+	self.toBigIcon:Destroy()
+	self.toNormalIcon:Destroy()
```
vocksel: Do Tweens actually have to be destroyed? Given that they're only stored in memory they should be cleaned up regardless when ProductItem is unmounted. They also don't seem like they have a Destroy method

docs/tutorials/shop-page.md
```diff
@@ -0,0 +1,396 @@
...
+In our case, the ShopPage component we're making will receive it's products from a parent component: that means that the product items will be received as props in our component.
+
+> Props or state?
+>
+> This is a question that often comes to our mind when we start working Roact. Remember that state is internal to the component. Things that are relevant only to a component should go in the state.
+>
+> On the other hand, props are used to pass data from parent components to child components. It implies that both parent component and it's child are aware of the structure of the props dictionary and depends on it.
+
+For our first draft of the component, we will define some props to be able to pass parameters to customize our shop page. We will start by adding the following:
+ - **frameProps:** a dictionary of props that will be passed to the scrolling frame to change it's visuals
```
vocksel: Could frameProps be removed? It seems strange to include as changing the style of the ScrollingFrame isn't part of what we want to be teaching. It seemed like you were modifying frameProps to change the CanvasSize but could that be handled differently without bloating the tutorial?

docs/tutorials/shop-page.md
jeparlefrancais: I don't think you can find the info on the [devhub](https://developer.roblox.com/en-us/api-reference/class/Tween). It does not show the ancestor functions and I don't really understand how I am suppose to know what are the ancestors of the class I am looking at. In fact, it's not really clear if it's an Instance

Anyway, you can check the ancestors on [rodocs](https://rodocs.github.io/mini/#Tween) and yes apparently it's a descendant of Instance.

But should we still destroy it? I wanted to have an example of clean up you can do in the `willUnmount` lifecycle method.

examples/shop-page/Components/ShopPage.lua
```diff
@@ -0,0 +1,73 @@
...
+local ShopPage = Roact.Component:extend("ShopPage")
+
+ShopPage.defaultProps = {
+	padding = 0,
+}
+
+function ShopPage:init()
+	self:setState({
+		cellSize = 100,
+		canvasHeight = 100,
```
MisterUncloaked: Since we don't know the actual cellSize or canvasHeight until the first call to onAbsoluteSizeChanged, should we default these to zero and not render anything in that case?

docs/tutorials/shop-page.md
```diff
@@ -0,0 +1,396 @@
...
+
+### Viewing the component
+
+Now that we have started working on our first component, we want to be able to see the results. To do that, let's create a LocalScript that will put out component in context. This script needs to _mount_ our ShopPage into a parent that will simply be a ScreenGui.
+
+```lua
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+local Players = game:GetService("Players")
+
+local Components = ReplicatedStorage:WaitForChild("Components")
```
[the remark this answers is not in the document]
jeparlefrancais: I added a comment above to try to make it clearer, if you think of anything else we could do let me know! Or you can submit a suggestion too!

examples/shop-page/Products.lua
```diff
@@ -0,0 +1,38 @@
+return {
+	{
+		identifier = "Red Visor",
+		image = "https://www.roblox.com/asset-thumbnail/image?assetId=139618072&width=420&height=420&format=png",
+		price = 30,
+		productId = 0,
```
MisterUncloaked: should we use real productIds and use those as the stable keys for the list items so we can get rid of the identifier field?

examples/shop-page/Components/ItemList.lua
```diff
@@ -0,0 +1,28 @@
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+
+local Components = script.Parent
+
+local Roact = require(ReplicatedStorage.Roact)
+
+local Item = require(Components:WaitForChild("Item"))
```
MisterUncloaked: Why do we need WaitForChild here? Is it unsafe for children of a folder to assume its siblings are available?

jeparlefrancais: I don't think there is any guarantee that all the siblings are replicated. From my experience it's best to use WaitForChild since you're sure everything is there when you need it.

Maybe @LPGhatguy can validate this!

examples/shop-page/Components/ShopPage.lua
jeparlefrancais: Sounds reasonable to me

examples/shop-page/Products.lua
jeparlefrancais: Hum I could put random values... Even if we keep the `0` values I could use the list index as a key and get rid of the identifier.

I think at first my plan was to show the name identifier in a label but I dropped that.

docs/tutorials/shop-page.md
jeparlefrancais: My goal was to make the frame style accessible from the top level component, so someone could easily modify it from the script used to mount the component to fit his style

Having it also allow to tell the reader that the props cannot be mutated in the ShopPage component.

How would you change it?

PR #223, opened 2019-07-02
Shallow wrapper

### LINKED ISSUE #189: Shallow Rendering
https://reactjs.org/docs/shallow-renderer.html

jeparlefrancais:
Closes #189.

So here is my first attempt at shallow rendering. For now it's more an illusion of shallow rendering because I have not touched at the renderer or the reconciler (yet).

How it works:
  1. It mounts the element in a host parent
  2. Then it wraps the VirtualNode in a ShallowWrapper
  3. That wrapper is used to make assertions on element kind, props or to find children of a component with some constraints (or simulate events later maybe)

This approach does not have much internal representation, it just uses the VirtualNode to find what it needs. It currently supports fragments.

The thing I'm probably the most unsure about is the notion of depth. You can specify the depth when calling `shallow(element, {depth = 1})`. I made a bunch of tests to document how it should behave in different situations but let me know if it's cohesive and make sense.

Let me know what you think about the approach in general.

Checklist before submitting:

### DIFF
--- a/src/shallow.lua
+++ b/src/shallow.lua
@@ -1,0 +1,128 @@
+local createReconciler = require(script.Parent.createReconciler)
+local RobloxRenderer = require(script.Parent.RobloxRenderer)
+local ElementKind = require(script.Parent.ElementKind)
+local ElementUtils = require(script.Parent.ElementUtils)
+
+local robloxReconciler = createReconciler(RobloxRenderer)
+
+local function getTypeFromVirtualNode(virtualNode)
+	local element = virtualNode.currentElement
+	local kind = ElementKind.of(element)
+
+	if kind == ElementKind.Host then
+		return {
+			kind = ElementKind.Host,
+			className = element.component,
+		}
+	elseif kind == ElementKind.Function then
+		return {
+			kind = ElementKind.Function,
+			functionComponent = element.component,
+		}
+	elseif kind == ElementKind.Stateful then
+		return {
+			kind = ElementKind.Stateful,
+			component = element.component,
+		}
+	else
+		error('>>> unkown element ' .. tostring(kind))
+	end
+end
+
+local ShallowWrapper = {}
+local ShallowWrapperMetatable = {
+	__index = ShallowWrapper,
+}
+
+function ShallowWrapper.new(virtualNode)
+	local wrapper = {
+		_virtualNode = virtualNode,
+		type = getTypeFromVirtualNode(virtualNode),
+		props = virtualNode.currentElement.props,
+	}
+
+	return setmetatable(wrapper, ShallowWrapperMetatable)
+end
+
+function ShallowWrapper:childrenCount()
+	local count = 0
+
+	for _ in pairs(self._virtualNode.children) do
+		count = count + 1
+	end
+
+	return count
+end
+
+function ShallowWrapper:find(constraints)
+	local results = {}
+
+	for _, child in pairs(self._virtualNode.children) do
+		if self:_satisfiesAllContraints(child, constraints) then
+			table.insert(results, ShallowWrapper.new(child))
+		end
+	end
+
+	return results
+end
+
+function ShallowWrapper:_satisfiesAllContraints(virtualNode, constraints)
+	for constraint, value in pairs(constraints) do
+		if not self:_satisfiesConstraint(virtualNode, constraint, value) then
+			return false
+		end
+	end
+
+	return true
+end
+
+function ShallowWrapper:_satisfiesConstraint(virtualNode, constraint, value)
+	local element = virtualNode.currentElement
+
+	if constraint == "kind" then
+		return ElementKind.of(element) == value
+
+	elseif constraint == "className" then
+		local isHost = ElementKind.of(element) == ElementKind.Host
+		return isHost and element.component == value
+
+	elseif constraint == "component" then
+		return element.component == value
+
+	elseif constraint == "props" then
+		local elementProps = element.props
+
+		for propKey, propValue in pairs(value) do
+			if elementProps[propKey] ~= propValue then
+				return false
+			end
+		end
+
+		return true
+	else
+		error(('unknown constraint %q'):format(constraint))
+	end
+end
+
+local function shallow(element, options)
+	options = options or {}
+
+	local tempParent = Instance.new("Folder")
+	local virtualNode = robloxReconciler.mountVirtualNode(element, tempParent, "ShallowTree")
+
+	local maxDepth = options.depth or 1
+	local currentDepth = 0
+
+	local nextNode = virtualNode
+	local wrapVirtualNode
+
+	repeat
+		wrapVirtualNode = nextNode
+		nextNode = wrapVirtualNode.children[ElementUtils.UseParentKey]
+		currentDepth = currentDepth + 1
+	until currentDepth > maxDepth or nextNode == nil
+
+	return ShallowWrapper.new(wrapVirtualNode)
+end
+
+return shallow
--- a/src/shallow.lua
+++ b/src/shallow.lua
@@ -25,18 +25,61 @@ local function getTypeFromVirtualNode(virtualNode)
 			component = element.component,
 		}
 	else
-		error('>>> unkown element ' .. tostring(kind))
+		error(('shallow wrapper does not support element of kind %q'):format(kind))
 	end
 end
+
+local function findNextVirtualNode(virtualNode, maxDepth)
+	local currentDepth = 0
+	local wrapVirtualNode = virtualNode
+	local nextNode = wrapVirtualNode.children[ElementUtils.UseParentKey]
+
+	while currentDepth < maxDepth and nextNode ~= nil do
+		wrapVirtualNode = nextNode
+		nextNode = wrapVirtualNode.children[ElementUtils.UseParentKey]
+		currentDepth = currentDepth + 1
+	end
+
+	return wrapVirtualNode
+end
+
+local ContraintFunctions = {
+	kind = function(virtualNode, expectKind)
+		return ElementKind.of(virtualNode.currentElement) == expectKind
+	end,
+	className = function(virtualNode, className)
+		local element = virtualNode.currentElement
+		local isHost = ElementKind.of(element) == ElementKind.Host
+		return isHost and element.component == className
+	end,
+	component = function(virtualNode, expectComponentValue)
+		return virtualNode.currentElement.component == expectComponentValue
+	end,
+	props = function(virtualNode, propSubSet)
+		local elementProps = virtualNode.currentElement.props
+
+		for propKey, propValue in pairs(propSubSet) do
+			if elementProps[propKey] ~= propValue then
+				return false
+			end
+		end
+
+		return true
+	end
+}
 
 local ShallowWrapper = {}
 local ShallowWrapperMetatable = {
 	__index = ShallowWrapper,
 }
 
-function ShallowWrapper.new(virtualNode)
+function ShallowWrapper.new(virtualNode, maxDepth)
+	virtualNode = findNextVirtualNode(virtualNode, maxDepth)
+
 	local wrapper = {
 		_virtualNode = virtualNode,
+		_childrenMaxDepth = maxDepth - 1,
+		_children = maxDepth == 0 and {} or virtualNode.children,
 		type = getTypeFromVirtualNode(virtualNode),
 		props = virtualNode.currentElement.props,
 	}
@@ -47,7 +90,7 @@ end
 function ShallowWrapper:childrenCount()
 	local count = 0
 
-	for _ in pairs(self._virtualNode.children) do
+	for _ in pairs(self._children) do
 		count = count + 1
 	end
 
@@ -57,9 +100,17 @@ end
 function ShallowWrapper:find(constraints)
 	local results = {}
 
-	for _, child in pairs(self._virtualNode.children) do
-		if self:_satisfiesAllContraints(child, constraints) then
-			table.insert(results, ShallowWrapper.new(child))
+	for constraint in pairs(constraints) do
+		if not ContraintFunctions[constraint] then
+			error(('unknown constraint %q'):format(constraint))
+		end
+	end
+
+	for _, child in pairs(self._children) do
+		local childWrapper = ShallowWrapper.new(child, self._childrenMaxDepth)
+
+		if self:_satisfiesAllContraints(childWrapper._virtualNode, constraints) then
+			table.insert(results, childWrapper)
 		end
 	end
 
@@ -68,40 +119,14 @@ end
 
 function ShallowWrapper:_satisfiesAllContraints(virtualNode, constraints)
 	for constraint, value in pairs(constraints) do
-		if not self:_satisfiesConstraint(virtualNode, constraint, value) then
+		local constraintFunction = ContraintFunctions[constraint]
+
+		if not constraintFunction(virtualNode, value) then
 			return false
 		end
 	end
 
 	return true
-end
-
-function ShallowWrapper:_satisfiesConstraint(virtualNode, constraint, value)
-	local element = virtualNode.currentElement
-
-	if constraint == "kind" then
-		return ElementKind.of(element) == value
-
-	elseif constraint == "className" then
-		local isHost = ElementKind.of(element) == ElementKind.Host
-		return isHost and element.component == value
-
-	elseif constraint == "component" then
-		return element.component == value
-
-	elseif constraint == "props" then
-		local elementProps = element.props
-
-		for propKey, propValue in pairs(value) do
-			if elementProps[propKey] ~= propValue then
-				return false
-			end
-		end
-
-		return true
-	else
-		error(('unknown constraint %q'):format(constraint))
-	end
 end
 
 local function shallow(element, options)
@@ -111,18 +136,8 @@ local function shallow(element, options)
 	local virtualNode = robloxReconciler.mountVirtualNode(element, tempParent, "ShallowTree")
 
 	local maxDepth = options.depth or 1
-	local currentDepth = 0
 
-	local nextNode = virtualNode
-	local wrapVirtualNode
-
-	repeat
-		wrapVirtualNode = nextNode
-		nextNode = wrapVirtualNode.children[ElementUtils.UseParentKey]
-		currentDepth = currentDepth + 1
-	until currentDepth > maxDepth or nextNode == nil
-
-	return ShallowWrapper.new(wrapVirtualNode)
+	return ShallowWrapper.new(virtualNode, maxDepth)
 end
 
 return shallow
--- a/src/shallow.lua
+++ b/src/shallow.lua
@@ -5,6 +5,11 @@ local ElementUtils = require(script.Parent.ElementUtils)
 
 local robloxReconciler = createReconciler(RobloxRenderer)
 
+local ShallowWrapper = {}
+local ShallowWrapperMetatable = {
+	__index = ShallowWrapper,
+}
+
 local function getTypeFromVirtualNode(virtualNode)
 	local element = virtualNode.currentElement
 	local kind = ElementKind.of(element)
@@ -25,7 +30,7 @@ local function getTypeFromVirtualNode(virtualNode)
 			component = element.component,
 		}
 	else
-		error(('shallow wrapper does not support element of kind %q'):format(kind))
+		error(('shallow wrapper does not support element of kind %q'):format(tostring(kind)))
 	end
 end
 
@@ -44,19 +49,18 @@ local function findNextVirtualNode(virtualNode, maxDepth)
 end
 
 local ContraintFunctions = {
-	kind = function(virtualNode, expectKind)
-		return ElementKind.of(virtualNode.currentElement) == expectKind
+	kind = function(element, expectKind)
+		return ElementKind.of(element) == expectKind
 	end,
-	className = function(virtualNode, className)
-		local element = virtualNode.currentElement
+	className = function(element, className)
 		local isHost = ElementKind.of(element) == ElementKind.Host
 		return isHost and element.component == className
 	end,
-	component = function(virtualNode, expectComponentValue)
-		return virtualNode.currentElement.component == expectComponentValue
+	component = function(element, expectComponentValue)
+		return element.component == expectComponentValue
 	end,
-	props = function(virtualNode, propSubSet)
-		local elementProps = virtualNode.currentElement.props
+	props = function(element, propSubSet)
+		local elementProps = element.props
 
 		for propKey, propValue in pairs(propSubSet) do
 			if elementProps[propKey] ~= propValue then
@@ -68,10 +72,36 @@ local ContraintFunctions = {
 	end
 }
 
-local ShallowWrapper = {}
-local ShallowWrapperMetatable = {
-	__index = ShallowWrapper,
-}
+local function countChildrenOfElement(element)
+	if ElementKind.of(element) == ElementKind.Fragment then
+		local count = 0
+
+		for _, subElement in pairs(element.elements) do
+			count = count + countChildrenOfElement(subElement)
+		end
+
+		return count
+	else
+		return 1
+	end
+end
+
+local function findChildren(virtualNode, constraints, results, maxDepth)
+	if ElementKind.of(virtualNode.currentElement) == ElementKind.Fragment then
+		for _, subVirtualNode in pairs(virtualNode.children) do
+			findChildren(subVirtualNode, constraints, results, maxDepth)
+		end
+	else
+		local childWrapper = ShallowWrapper.new(
+			virtualNode,
+			maxDepth
+		)
+
+		if childWrapper:_satisfiesAllContraints(constraints) then
+			table.insert(results, childWrapper)
+		end
+	end
+end
 
 function ShallowWrapper.new(virtualNode, maxDepth)
 	virtualNode = findNextVirtualNode(virtualNode, maxDepth)
@@ -90,38 +120,37 @@ end
 function ShallowWrapper:childrenCount()
 	local count = 0
 
-	for _ in pairs(self._children) do
-		count = count + 1
+	for _, virtualNode in pairs(self._children) do
+		local element = virtualNode.currentElement
+		count = count + countChildrenOfElement(element)
 	end
 
 	return count
 end
 
 function ShallowWrapper:find(constraints)
-	local results = {}
-
 	for constraint in pairs(constraints) do
 		if not ContraintFunctions[constraint] then
 			error(('unknown constraint %q'):format(constraint))
 		end
 	end
 
-	for _, child in pairs(self._children) do
-		local childWrapper = ShallowWrapper.new(child, self._childrenMaxDepth)
+	local results = {}
 
-		if self:_satisfiesAllContraints(childWrapper._virtualNode, constraints) then
-			table.insert(results, childWrapper)
-		end
+	for _, childVirtualNode in pairs(self._children) do
+		findChildren(childVirtualNode, constraints, results, self._childrenMaxDepth)
 	end
 
 	return results
 end
 
-function ShallowWrapper:_satisfiesAllContraints(virtualNode, constraints)
+function ShallowWrapper:_satisfiesAllContraints(constraints)
+	local element = self._virtualNode.currentElement
+
 	for constraint, value in pairs(constraints) do
 		local constraintFunction = ContraintFunctions[constraint]
 
-		if not constraintFunction(virtualNode, value) then
+		if not constraintFunction(element, value) then
 			return false
 		end
 	end
--- a/src/shallow.spec.lua
+++ b/src/shallow.spec.lua
@@ -162,6 +162,57 @@ return function()
 			})
 			expect(#childB).to.equal(1)
 		end)
+
+		it("should unwrap children when depth is two", function()
+			local element = createElement(ComponentWithChildren)
+
+			local result = shallow(element, {
+				depth = 2,
+			})
+
+			local hostChild = result:find({
+				component = unwrappedClassName,
+			})
+			expect(#hostChild).to.equal(1)
+
+			local unwrappedBChild = result:find({
+				component = A,
+			})
+			expect(#unwrappedBChild).to.equal(1)
+		end)
+
+		it("should not include any children when depth is zero", function()
+			local element = createElement(ComponentWithChildren)
+
+			local result = shallow(element, {
+				depth = 0,
+			})
+
+			expect(result:childrenCount()).to.equal(0)
+		end)
+
+		it("should not include any grand-children when depth is one", function()
+			local function ParentComponent()
+				return createElement("Folder", {}, {
+					Child = createElement(ComponentWithChildren),
+				})
+			end
+
+			local element = createElement(ParentComponent)
+
+			local result = shallow(element, {
+				depth = 1,
+			})
+
+			expect(result:childrenCount()).to.equal(1)
+
+			local componentWithChildrenWrapper = result:find({
+				component = ComponentWithChildren,
+			})[1]
+			expect(componentWithChildrenWrapper).to.be.ok()
+
+			expect(componentWithChildrenWrapper:childrenCount()).to.equal(0)
+		end)
 	end)
 
 	describe("childrenCount", function()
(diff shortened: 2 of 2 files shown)


jeparlefrancais:
So this just turned into a massive PR, more oriented towards generating snapshots.

The ShallowWrapper will need a better name, but I'm still not sure about what it should be (open for suggestions!)

The current API for snapshots is 
```lua
local wrapper = Roact.shallow(element, {
    depth = 1,
})

wrapper:toMatchSnapshot("snapshot identifier")
```
This would generate a snapshot serialized in a `ModuleScript` in a folder named "RoactSnapshots" under `ReplicatedStorage`

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+		local childWrapper = children[i]
+		if childWrapper:_satisfiesAllContraints(constraints) then
+			table.insert(results, childWrapper)
+		end
+	end
+
+	return results
+end
+
+function ShallowWrapper:findUnique(constraints)
```
MisterUncloaked: Do we need convenience methods like this? I'm trying to imagine when developers will want to use this versus relying on snapshot matching.

MisterUncloaked:
Overall I really like the idea of snapshotting the shallow representation of components versus how snapshots are traditionally done. It keeps the snapshots smaller, and ultimately removes the need for a lot of boilerplate assertions typically found when testing with shallow rendering.

src/snapshot.spec.lua
```diff
@@ -0,0 +1,99 @@
...
+	it("should match snapshot of stateful component", function()
+		local StatefulComponent = Component:extend("CoolComponent")
+
+		function StatefulComponent:render()
+			return createElement("TextLabel")
+		end
+
+		local element = createElement("Frame", {}, {
+			Child = createElement(StatefulComponent, {
+				label = {
```
MisterUncloaked: label shows up in my snapshot as label = Markers.Unknown
Do we want to support props that are tables?

src/VirtualTree.lua
```diff
@@ -0,0 +1,33 @@
...
+function VirtualTree.new(rootNode, internalDataSymbol)
+	local tree = {
+		[Type] = Type.VirtualTree,
+		[internalDataSymbol] = {
+			rootNode = rootNode,
+			mounted = true,
+		}
+	}
+
+	function tree:getTestRenderOutput(options)
```
MisterUncloaked: for a V1 we might want to hide shallows as an implementation detail. We can have a single API that will be really easy to use in tests, for example:

`
local tree = Roact.mount(element)
Roact.matchSnapshot(tree)
Roact.unmount(tree)`

src/assertDeepEqual.lua
```diff
@@ -5,56 +5,7 @@
...
-	end
-
-	if a == b then
-		return true
-	end
-
-	local message = "{1} ~= {2}"
-	return false, message
-end
+local deepEqual = require(script.Parent.deepEqual)
```
MisterUncloaked: nice refactor

src/VirtualTree.lua
```diff
@@ -0,0 +1,33 @@
...
+local config = require(script.Parent.GlobalConfig).get()
+
+local VirtualTree = {}
+
+function VirtualTree.new(rootNode, internalDataSymbol)
+	local tree = {
+		[Type] = Type.VirtualTree,
+		[internalDataSymbol] = {
+			rootNode = rootNode,
+			mounted = true,
```
MisterUncloaked: we should pass mounted as a parameter instead of assuming a tree is always created as mounted (though that's always true right now)

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+			kind = ElementKind.Function,
+			functionComponent = element.component,
+		}
+	elseif kind == ElementKind.Stateful then
+		return {
+			kind = ElementKind.Stateful,
+			component = element.component,
+		}
+	else
+		error(('shallow wrapper does not support element of kind %q'):format(tostring(kind)))
```
MisterUncloaked: reminder for us to document the types we support when we get to writing documentation for snapshot testing

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+		end
+	end
+
+	return filteredProps
+end
+
+function ShallowWrapper.new(virtualNode, maxDepth)
+	virtualNode = findNextVirtualNode(virtualNode, maxDepth)
+
+	local wrapper = {
```
MisterUncloaked: to simplify this class, the wrapper could just be the node + the depth, and we could remove some of the mapping we are doing from the virtual node interface to the shallow

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+		local childWrapper = ShallowWrapper.new(
+			virtualNode,
+			maxDepth
+		)
+
+		table.insert(results, childWrapper)
+	end
+end
+
+local function filterProps(props)
```
MisterUncloaked: it could be simpler to just not filter the props in the shallow at all

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+		#constrainedChildren == 1,
+		("expect to find only one child, but found %d"):format(#constrainedChildren)
+	)
+
+	return constrainedChildren[1]
+end
+
+function ShallowWrapper:getChildren()
+	if self._shallowChildren then
+		return self._shallowChildren
```
MisterUncloaked: could be unsafe to do this kind of caching - I think the less state the shallow has the better chance it has of working at any point in time on the underlying tree

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+	end
+
+	local results = {}
+
+	for _, childVirtualNode in pairs(self._virtualNodeChildren) do
+		getChildren(childVirtualNode, results, self._childrenMaxDepth)
```
MisterUncloaked: instead of using a non member function, this could be a bit simpler:

```
local childWrapper = ShallowWrapper.new(
	childVirtualNode,
	self._childrenMaxDepth
)
childWrapper:getChildren()
table.insert(results, childWrapper)
```

this would need the fragment check as well if we got rid of the other function

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+	end
+
+	self._shallowChildren = results
+	return results
+end
+
+function ShallowWrapper:toMatchSnapshot(identifier)
```
MisterUncloaked: I think it makes sense to keep the idea of snapshots separate from the idea of shallows - this API can be the public function I was describing in another comment

src/snapshot/Serialize/SnapshotData.lua
```diff
@@ -0,0 +1,120 @@
...
+local Type = require(RoactRoot.Type)
+local Ref = require(RoactRoot.PropMarkers.Ref)
+
+local function sortSerializedChildren(childA, childB)
+	return childA.hostKey < childB.hostKey
+end
+
+local SnapshotData = {}
+
+function SnapshotData.type(wrapperType)
```
MisterUncloaked: I wonder if SnapshotData and Serializer could be the same thing - a single source of truth to go from Shallow to a string representing a ModuleScript (snapshot)

src/snapshot/Snapshot.lua
```diff
@@ -0,0 +1,101 @@
...
+	if not (existingData and existingData:IsA('StringValue')) then
+		existingData = Instance.new("StringValue")
+		existingData.Name = self._identifier
+		existingData.Parent = folder
+	end
+
+	existingData.Value = snapshotSource
+end
+
+function Snapshot.getSnapshotFolder()
```
MisterUncloaked: would be nice to give developers control over this - maybe an option to pass in Roact.matchSnapshot

src/snapshot/Serialize/SnapshotData.lua
jeparlefrancais: I am not sure if we can do that because when we compare with the generated snapshot with the previous serialized one, we use that intermediate representation to do the comparison.

So I think we're kind of stuck with that intermediate representation! If you have any idea of how we could do it though let me know, because it would reduce one layer of complexity so that would be great

MisterUncloaked: Another alternative is to rename SnapshotData to Snapshot, and then Snapshot could be called SnapshotMatcher. It is nice to keep them separate from Serializer so you have a single entity to call deepEqual on as you have now

src/ShallowWrapper.lua
```diff
@@ -0,0 +1,228 @@
...
+		return {
+			kind = ElementKind.Stateful,
+			component = element.component,
+		}
+	else
+		error(('shallow wrapper does not support element of kind %q'):format(tostring(kind)))
+	end
+end
+
+local function findNextVirtualNode(virtualNode, maxDepth)
```
MisterUncloaked: for my own knowledge, what's the role of this function? If we call Shallow.new why can't we assume we passed in the next virtual node?

jeparlefrancais: That is the thing that gives the illusion of shallow rendering, I have not documented it a lot but if you take a look at the notes repo, checkout the tips page there is a example when you are using `depth = 2`.

So `findNextVirtualNode` will get the next virtual node that wraps a host element (a roblox instance). When you allow more _depth_ to the shallow wrapper, it tries to render to roblox instances as much as it can.

I have written a lot of tests to verify the behavior of depth, it's a bit confusing honestly. I've changed the definition a few times before getting this version. It might be something worth reviewing together too if you want

MisterUncloaked: Depth might be more intuitive if it still preserves the parent. I tried running some of the tests like stateful-component-children with a depth of 2, and I would have expected CoolComponent to still be in my snapshot, but get to see its children.

src/ShallowWrapper.lua
jeparlefrancais: The thing I don't like with the member function is that I would have to copy back the elements from the `getChildren` call into the result table recursively. Do you think it would be better to just go for clarity at this point? I have not benchmark anything yet so I don't have numbers about what is the performance gain

MisterUncloaked: Perf is a good point, but I wonder if that concern goes away if we redefine depth such that we're building more deeply nested tables instead of coalescing the children at various levels of the tree to be the shallow children.

src/snapshot.spec.lua
jeparlefrancais: Fixed!

src/ShallowWrapper.lua
jeparlefrancais: I want remove the Roact.Children key from the props when it's there. That is because it makes it easier to assert your props. Intuitively, it matches what you put in the props argument of Roact.createElement.

src/ShallowWrapper.lua
MisterUncloaked: We decided to keep this for cases where we want to remove some scaffolding injected by things like mock providers

src/snapshot/Serialize/Snapshot.lua
```diff
@@ -0,0 +1,123 @@
...
+
+		serializedChildren[i] = Snapshot.wrapper(childWrapper)
+	end
+
+	table.sort(serializedChildren, sortSerializedChildren)
+
+	return serializedChildren
+end
+
+function Snapshot.wrapper(wrapper)
```
MisterUncloaked: should this be called Snapshot.new?

src/snapshot/SnapshotMatcher.lua
```diff
@@ -0,0 +1,101 @@
+local ReplicatedStorage = game:GetService("ReplicatedStorage")
+
+local Markers = require(script.Parent.Serialize.Markers)
+local Serialize = require(script.Parent.Serialize)
+local deepEqual = require(script.Parent.Parent.deepEqual)
+local ElementKind = require(script.Parent.Parent.ElementKind)
+
+local SnapshotFolderName = "RoactSnapshots"
```
MisterUncloaked: minor: should be loud snake case SNAPSHOT_FOLDER_NAME for this local constant

src/snapshot/SnapshotMatcher.lua
```diff
@@ -0,0 +1,101 @@
...
+local SnapshotFolder = ReplicatedStorage:FindFirstChild(SnapshotFolderName)
+
+local SnapshotMatcher = {}
+local SnapshotMetatable = {
+	__index = SnapshotMatcher,
+	__tostring = function(snapshot)
```
MisterUncloaked: should Snapshot.lua return tables with a metatable like this instead of the matcher? It feels more intuitive to be able to call tostring on the snapshot itself instead of the matcher

jeparlefrancais: Yes I removed it and I added a `getSnapshotString` to ShallowWrapper and SnapshotMatcher

src/ShallowWrapper.lua
jeparlefrancais: I just removed the caching now

src/snapshot/SnapshotMatcher.lua
jeparlefrancais: should I make it configurable? or do we push that to another PR?

src/shallow/ShallowWrapper.lua
```diff
@@ -0,0 +1,232 @@
...
+
+function ShallowWrapper:matchSnapshot(identifier)
+	assert(typeof(identifier) == "string", "Snapshot identifier must be a string")
+
+	local snapshotResult = snapshot(identifier, self)
+
+	snapshotResult:match()
+end
+
+function ShallowWrapper:getSnapshotString(identifier)
```
MisterUncloaked: I think we should support getting the string without needing to pass an identifier or deal with existing data. This could be something like

```
return Serialize.snapshotToString(Snapshot.new(self))

File: src/shallow/snapshot/Serialize/init.lua [hunk 19]
```diff
@@ -0,0 +1,11 @@
+local Serializer = require(script.Serializer)
+local Snapshot = require(script.Snapshot)
+
+return {
+	wrapperToSnapshot = function(wrapper)
+		return Snapshot.new(wrapper)
+	end,
+	snapshotToString = function(snapshot)
+		return Serializer.firstSnapshotData(snapshot)
```
MisterUncloaked: For consistency, we should just call this snapshotToString too

src/shallow/ShallowWrapper.lua
```diff
@@ -0,0 +1,232 @@
...
+	assert(typeof(identifier) == "string", "Snapshot identifier must be a string")
+
+	local snapshotResult = snapshot(identifier, self)
+
+	return snapshotResult:getSnapshotString()
+end
+
+function ShallowWrapper:_satisfiesAllContraints(constraints)
```
MisterUncloaked: this is a good candidate to not be a member function to make it truly private. Consumers will get access to ShallowWrapper directly, so we should hide as much as we can versus using the underscore naming convention

MisterUncloaked: It could also make sense to put all these utility functions into their own file so it's super clear that everything in ShallowWrapper is available to consumers

src/shallow/snapshot/SnapshotMatcher.lua
```diff
@@ -0,0 +1,103 @@
...
+	if not SnapshotFolder then
+		SnapshotFolder = Instance.new("Folder")
+		SnapshotFolder.Name = SnapshotFolderName
+		SnapshotFolder.Parent = ReplicatedStorage
+	end
+
+	return SnapshotFolder
+end
+
+function SnapshotMatcher._loadExistingData(identifier)
```
MisterUncloaked: another candidate for a local function to guarantee privacy since we return the ShallowMatcher up to the ShallowWrapper (though it's not exposed to Roact consumers)

src/shallow/snapshot/Serialize/Serializer.lua
```diff
@@ -0,0 +1,235 @@
...
+	Serializer.snapshotDataContent(snapshotData, output)
+	output:popAndWrite("},")
+end
+
+function Serializer.firstSnapshotData(snapshotData)
+	local output = IndentedOutput.new()
+	output:writeAndPush("return function(dependencies)")
+	output:write("local Roact = dependencies.Roact")
+	output:write("local ElementKind = dependencies.ElementKind")
+	output:write("local Markers = dependencies.Markers")
```
MisterUncloaked: Since folks will be committing snapshots, I feel like we should make sure the lua files we generate pass luacheck. We could add the ability to mark dependencies as we see them in the output, though it would require reworking this file a bit.

jeparlefrancais: What do we do for the indentation? Right now I indent snapshot files with two spaces to make it more readable, but usually we indent with tabs.  I don't know what we gain from passing luacheck on a generated file too.

Right now there are only warnings showing up

src/shallow/init.lua
```diff
@@ -0,0 +1,21 @@
...
+
+local shallowTreeKey = "RoactTree"
+
+local function shallow(element, options)
+	assert(Type.of(element) == Type.Element, "Expected arg #1 to be an Element")
+
+	options = options or {}
+	local maxDepth = options.depth or 1
+
+	local rootNode = robloxReconciler.mountVirtualNode(element, nil, shallowTreeKey)
```
MisterUncloaked: I'm imagining most tests will be inserting the shallow logic in between calls to mount and unmount. Because of this, we should just take the tree as an argument, and figure out how to get access to the root node

src/shallow/snapshot/SnapshotMatcher.lua
jeparlefrancais: The problem is that I can not mock it when I want to write tests! What should I do?

MisterUncloaked: That's true. We can leave it for now since it's not part of the public API

src/shallow/snapshot/Serialize/Serializer.lua
MisterUncloaked: That's fair - we can add to the documentation a reminder not to luacheck these. I figured developers tend to luacheck all their checked in lua files, but maybe that is not the case.

src/snapshot/SnapshotMatcher.lua
MisterUncloaked: Doesn't need to be configurable right now

src/shallow/ShallowWrapper.lua
```diff
@@ -0,0 +1,202 @@
...
+function ShallowWrapperPublic:matchSnapshot(identifier)
+	assert(typeof(identifier) == "string", "Snapshot identifier must be a string")
+
+	local snapshotResult = Snapshot.createMatcher(identifier, self)
+
+	snapshotResult:match()
+end
+
+function ShallowWrapperPublic:snapshotToString()
+	return Snapshot.toString(self)
```
MisterUncloaked: should we just have a __tostring meta method for this on ShallowWrapper?

src/init.lua
```diff
@@ -3,14 +3,19 @@
...
 local createReconcilerCompat = require(script.createReconcilerCompat)
-local RobloxRenderer = require(script.RobloxRenderer)
 local strict = require(script.strict)
 local Binding = require(script.Binding)
+local VirtualTree = require(script.VirtualTree)
 
-local robloxReconciler = createReconciler(RobloxRenderer)
-local reconcilerCompat = createReconcilerCompat(robloxReconciler)
+local function mount(element, hostParent, hostKey)
+	return VirtualTree.mount(element, {
```
MisterUncloaked: should VirtualTree.mount have the signature `mount(element, hostParent, hostKey)` so we don't have to declare a local function here?

src/VirtualTree.lua
```diff
@@ -0,0 +1,86 @@
+local createReconciler = require(script.Parent.createReconciler)
+local RobloxRenderer = require(script.Parent.RobloxRenderer)
+local shallow = require(script.Parent.shallow)
+local Symbol = require(script.Parent.Symbol)
+local Type = require(script.Parent.Type)
+
+local config = require(script.Parent.GlobalConfig).get()
+
+local DEFAULT_RENDERER = createReconciler(RobloxRenderer)
```
MisterUncloaked: should the reconciler be passed to the tree instead of having all mount calls have to use the same reconciler?

docs/api-reference.md
```diff
@@ -608,4 +624,140 @@ end
...
+---
+
+## ShallowWrapper
+
+### Fields
+
+#### type
+```
+type: {
+	kind: ElementKind
```
MisterUncloaked: like with https://github.com/Roblox/roact/issues/227 we might want a mirror and not expose the ElementKind symbols directly, so Elements cannot be impersonated.

docs/api-reference.md
```diff
@@ -608,4 +624,140 @@ end
...
+```
+getChildren() -> list[ShallowWrapper]
+```
+Returns a list of all children wrapped into ShallowWrappers.
+
+---
+
+#### getInstance
+```
+getInstance() -> Instance or nil
```
MisterUncloaked: Do we need getInstance? tostring + matchsnapshot + supporting methods to query the public fields of the shallow tree should be enough

docs/api-reference.md
```diff
@@ -608,4 +624,140 @@ end
...
+Returns the string source of the snapshot. Useful for debugging purposes.
+
+---
+
+##### Constraints
+Constraints are passed through a dictionary that maps a constraint name to it's value.
+
+| name | value type | description |
+| --- | --- | --- |
+| kind | ElementKind | Filters with the ElementKind of the rendered elements |
```
MisterUncloaked: do we need API references in this file for ElementKind since we are exposing it now as public for these constraint functions?

docs/api-reference.md
```diff
@@ -44,6 +44,22 @@ The result is a `RoactTree`, which is an opaque handle that represents a tree of
 
 ---
 
+### Roact.shallow
```
MisterUncloaked: When we do the full markdown page on snapshot testing, we should walk through the use cases for when developers should use which methods of shallow wrapper. When should developers assert facts w/ calls to `find` versus only doing a snapshot?

docs/api-reference.md
```diff
@@ -608,4 +624,140 @@ end
...
+| Function | functionComponent: function | the function that renders the element |
+| Stateful | component: table | the class-like table used to render the element |
+
+---
+
+#### props
+The props of the ShallowWrapper.
+
+!!! note
+	This dictionary will not contain the `Roact.Children` prop. To obtain the children elements wrapped into a ShallowWrapper, use the method `getChildren()`
```
MisterUncloaked: Would it be simpler to just populate children as a property of ShallowWrapper when we call new? I'm trying to think of when you want a ShallowWrapper but wouldn't call getChildren()

src/init.lua
jeparlefrancais: I wanted `VirtualTree.mout` to accept a reconciler so it can be changed, but I did not wanted to have a 4 arguments function with three optional ones. It makes weird call when you want to pass only the reconciler (`VirtualTree.mount(element, nil, nil, reconciler)`). If one day mount needs a new argument (if we make a true shallow renderer for example), it's going to get worse.

So I decided to use a table to make it easier in the future if we add/remove other optional arguments.

src/shallow/ShallowWrapper.lua
jeparlefrancais: Hum I am not sure about `tostring(ShallowWrapper)`... It not very intuitive I think, because I would expect to receive a string version of a shallow wrapper instead of a generated snapshot from it. What do you think?

src/VirtualTree.lua
jeparlefrancais: Yes the reconciler is passed through the `options` table. My comment above gives more detail about that

docs/api-reference.md
jeparlefrancais: Yes if we don't expose ElementKind we should remove that!

docs/api-reference.md
jeparlefrancais: For more robust tests it can be interesting to directly assert on the instance. For example you can assert for a specific value of AbsoluteSize or AbsolutePosition. It could also useful if you want to test behavior from lifecycle methods that act on a ref.

src/shallow/ShallowWrapper.lua
MisterUncloaked: I was trying to think of what to name this, similar to react-test-renderer toJSON(). Maybe toSnapshotString is a bit better name and ya reserve tostring(ShallowWrapper) for a future string representation that is more focused on the shallow's structure.

docs/api-reference.md
MisterUncloaked: In such a test, could you just use a ref?

jeparlefrancais: If there is already a Ref, yes because you can access it through the props. Otherwise I don't think so

docs/api-reference.md
jeparlefrancais: Should I just make a `children` field and remove `getChildren` and `getChildrenCount`?

src/shallow/validateShallowOptions.lua
```diff
@@ -0,0 +1,30 @@
...
+			return false, ("unexpected option type for %q (expected %s but got %s)"):format(
+				tostring(key),
+				expectType,
+				typeof(value)
+			)
+		end
+	end
+
+	return true
+end
```
ZoteTheMighty: This solution seems way overzealous for something like this. Is this in the name of future proofing? I worry it's getting a little out of hand.

docs/api-reference.md
```diff
@@ -608,4 +608,149 @@ end
...
+}
+```
+
+The type dictionary always has the `kind` field that tell the component type. Additionally, depending of the kind of component, other information can be included.
+
+| kind | fields | description |
+| --- | --- | --- |
+| Host | className: string | the ClassName of the instance |
+| Function | functionComponent: function | the function that renders the element |
+| Stateful | component: table | the class-like table used to render the element |
```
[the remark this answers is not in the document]
jeparlefrancais: Yes I changed it while you were reviewing to remove the ElementKind field from the ShallowWrapper.

I replaced it with a single `component` field

docs/api-reference.md
```diff
@@ -32,21 +32,21 @@ Creates a new Roact fragment with the provided table of elements. Fragments allo
 
 ### Roact.mount
 ```
-Roact.mount(element, [parent, [key]]) -> RoactTree
+Roact.mount(element, [parent, [key]]) -> VirtualTree
```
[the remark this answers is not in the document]
jeparlefrancais: I can switch it back to RoactTree, it's just that the Type we are going to expose in #230 is called `VirtualTree`. So maybe it could be renamed in that other PR?

docs/scripts/sync-snapshots-with-lua.md
```diff
@@ -0,0 +1,102 @@
...
+
+print(filteredOutput, "\n")
+
+for snapshotPattern in output:gmatch("Snapshot:::<|[%w_%-%.]+|><|=>.-<=|>") do
+	local name = snapshotPattern:match("Snapshot:::<|([%w_%-%.]+)|>")
+	local content = snapshotPattern:match("<|=>(.-)<=|>")
+
+	writeSnapshotFile(name, content)
+end
+```
```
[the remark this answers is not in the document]
jeparlefrancais: I made both originally because with Roact, lemur runs with lua so I assumed people would prefer lua to sync stuff. And I have made the python version because uiblox is already using python for it's image packaging thing!

So I don't know if we need both, but it's more convenient to offer it I think

src/shallow/Snapshot/Serialize/Serializer.lua
```diff
@@ -0,0 +1,247 @@
...
+	if kind == ElementKind.Host then
+		return "Host"
+	elseif kind == ElementKind.Function then
+		return "Function"
+	elseif kind == ElementKind.Stateful then
+		return "Stateful"
+	else
+		error(("Cannot serialize ElementKind %q"):format(tostring(kind)))
+	end
+end
```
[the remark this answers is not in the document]
jeparlefrancais: Yes, I removed it from the ShallowWrapper object. Before you could access it with
```lua
wrapper.type.kind
```
The type field would also contain `className`, `functionComponent` or `classComponent` depending on the kind of the component.

Now, I removed it and simply offer a `component` field that is just like the component field from the VirtualNode. Which is a string for host's class name, a function or a table for function or stateful components.

When I create the SnapshotData from the ShallowWrapper, I get back the ElementKind from `ElementKind.fromComponent` in order to serialize it into the ModuleScript. So that way, the source looks like this
```lua
type = {
    kind = ElementKind.Stateful,
    componentName = "CoolComponent",
}
type = {
    kind = ElementKind.Function,
},
type = {
    kind = ElementKind.Host,
    className = "Frame",
},
```

As you can see, it's mostly useful for function components, because I don't have anything meaningful to serialize about their nature than their ElementKind.

src/shallow/Snapshot/Serialize/Snapshot.lua
```diff
@@ -0,0 +1,133 @@
...
+	return childA.hostKey < childB.hostKey
+end
+
+local Snapshot = {}
+
+function Snapshot.type(wrapperComponent)
+	local kind = ElementKind.fromComponent(wrapperComponent)
+
+	local typeData = {
+		kind = kind,
```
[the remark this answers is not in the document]
jeparlefrancais: I made a longer comment above, but it's useful for function components, because I can't serialize their function. It's only going to show up as a AnonymousFunction symbol

src/shallow/validateShallowOptions.lua
jeparlefrancais: Yes at some point I add three parameters (depth, hostKey and hostParent) so I felt like it would be good to make sure there are no typos and mismatched types. But yeah right now with only the depth it looks a bit like nuking a fly.

docs/api-reference.md
```diff
@@ -608,4 +608,149 @@ end
...
+end
+
+local function MainComponent(props)
+	return Roact.createElement("Frame", {}, {
+		Cool = Roact.createElement(CoolComponent),
+	})
+end
+```
+
+If we shallow-render the `MainComponent`, the default depth will not render the CoolComponent.
```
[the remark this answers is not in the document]
jeparlefrancais: I wanted to explain when `getHostObject` can return `nil` or not, but yep it does include talking a bit about depth... Should I remove it or not?

src/shallow/ShallowWrapper.lua
```diff
@@ -0,0 +1,159 @@
...
+	while currentDepth < maxDepth and nextNode ~= nil do
+		currentNode = nextNode
+		nextNode = currentNode.children[ElementUtils.UseParentKey]
+		currentDepth = currentDepth + 1
+	end
+
+	return currentNode
+end
+
+local function getChildren(virtualNode, results, maxDepth)
```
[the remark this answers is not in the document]
jeparlefrancais: I wanted to avoid copying table here because of the recursion. The way it's working right now, each child will add itself to the list, so it means that if make tables immutable I would need to create at least one list per child and recombine them. So are you sure it would be worth it to make it immutable here?

src/shallow/Snapshot/Serialize/IndentedOutput.lua
```diff
@@ -0,0 +1,60 @@
...
+	return output
+end
+
+function IndentedOutput:write(line, ...)
+	if select("#", ...) > 0 then
+		line = line:format(...)
+	end
+
+	if self._removeTrailingWhitespaces and line == "" then
+		table.insert(self._lines, line)
```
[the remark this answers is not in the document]
jeparlefrancais: Yes you made me realize it should not be called that I think. Because I am not sure it's worth checking for all trailing whitespaces, instead we can just assume with not going to write a line that ends with spaces.

But the thing that the consumer of IndentedOutput does not control is when it writes an empty line in order to skip a line. Then if there is an indentation level, I don't want to indent the empty string with spaces, I just want to add it so it creates a new line when concatenated.

src/shallow/validateShallowOptions.lua
jeparlefrancais: Should I still keep the dictionary as a parameter for future proofing, but remove the validation?

docs/api-reference.md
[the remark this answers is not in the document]
ZoteTheMighty: That's a good question! I didn't realize we were exposing it in that PR. We should probably discuss this with @LPGhatguy.

My thoughts are still that `VirtualTree` probably means less to users, but maybe there's another better alternative too.

src/shallow/Snapshot/Serialize/IndentedOutput.lua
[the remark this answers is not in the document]
ZoteTheMighty: Yeah that's definitely something else altogether.

Is this just for adding spacing between the requires and the returned table? It seems like a pretty specific edge case, but I think it might be reasonable to _always_ avoid indenting empty lines.

src/shallow/validateShallowOptions.lua
ZoteTheMighty: I think when it comes to stuff like this, we need to strike a balance between predicting how our API might change in the future and just giving up and making it a table. Tables are fine in a lot of cases but they feel really awkward when there are only one or two elements.

Personally, I'm okay with the public `shallow` function just taking depth as its second argument. It doesn't seem like we would wildly change this function's signature anytime soon. What do you think?

docs/api-reference.md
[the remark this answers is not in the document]
jeparlefrancais: Yes I agree that the term VirtualTree might be questionable to users, we should probably solves this in #230 if it's about to be merged first.

src/shallow/validateShallowOptions.lua
jeparlefrancais: Yes and you make me realise that since we are now mounting with `Roact.mount`, the shallow function is not public anymore. I use it as en entry point from Roact to the shallow module.

src/shallow/Snapshot/Serialize/IndentedOutput.lua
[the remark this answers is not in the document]
jeparlefrancais: Yes otherwise it's a bit annoying when you open the snapshot file with an editor, it removes the indentation from the empty line so then if it's already committed it appears in the diff. It's a bit of 
 a detail but I think it's worth it

PR #226, opened 2019-08-01
Fix benchmarks

MisterUncloaked:
## Problem

Benchmarks could neither be run from the command line nor from the place generated from place.project.json

1) Habitat:loadFromFs does not know to create a module script when loading a folder with an init.server.lua in it
2) Each benchmark assumed it lived relative to Roact which is not necessarily true - in place.project.json the benchmarks live in ServerScriptService so that that are run automatically

## Solution
1) Load the init file directly so that we get a module script to require, and not a folder
2) Always look to replicated storage when looking for Roact in the context of a benchmark

## Testing

Observed that benchmarks ran via

lua bin/bench.lua
rojo build place.project.json -o RoactTestPlace.rbxlx --> open and click play

### DIFF
--- a/benchmarks.project.json
+++ b/benchmarks.project.json
@@ -1,5 +1,5 @@
 {
-  "name": "Roact Test Place",
+  "name": "Roact Benchmarks Place",
   "tree": {
     "$className": "DataModel",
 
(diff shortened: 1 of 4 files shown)


jeparlefrancais:
The only thing I have to say is that I am not sure if running it by default with the place project is a good idea. It depends what the place file is supposed to be used for. If it is to show the different examples, running the benchmarks is not a good idea because when you press play, the window just freeze for a bit before showing anything.

Maybe we could add a benchmarks project file? Another solution could be to only have a script under `bin` that does the benchmark and use it with `run-in-roblox`. That way, we could run something like `run-in-roblox place.rbxlx -s bin/benchmarks.lua` to get the output.

benchmarks.project.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "Roact Benchmarks Place",
```
MisterUncloaked: @jeparlefrancais I liked your idea of keeping benchmarks in a separate place. Updated here.

Issue #227, opened 2019-08-07
Type Checking API


LPGhatguy:
It'd be useful to have a method to determine the type of objects created by Roact in order to validate them more rigorously.

We can expose this through a feature we already have internally, `Type`. The public API would be:

* `Roact.typeOf(object) -> RoactType | nil`
* `Roact.Type`, an enum with these variants, taken from [Type.lua](https://github.com/Roblox/roact/blob/9493406f0a4892b55dff98386c1d891458f5975d/src/Type.lua):
	* Binding
	* Element
	* HostChangeEvent
	* HostEvent
	* StatefulComponentClass
	* StatefulComponentInstance
	* VirtualNode -- objects of this type aren't currently exposed to users
	* VirtualTree

We should be careful not to expose the actual `Type` symbol that we use as a key for these types, which would allow users to impersonate Roact instances.

Alternatively, we could have this function return strings instead of enum variants. This would let us keep the set of types we expose open (potentially aiding modularity), but introduce the potential for typos from our users. This has precedent in Lua's `type` and Roblox's `typeof` functions.

MisterUncloaked:
I made some good progress on this today and will have a PR up tomorrow. I'm just working on documentation at this point.

PR #230, opened 2019-08-09
Add Type Checking API

MisterUncloaked:
Closes https://github.com/Roblox/roact/issues/227

Added a public API for type checking in Roact

- Added nameOf function to Type to lookup the string name for any Type
- Added TypeMirror module which exposes a subset of types in Type without exposing the internal Roact symbols
- Added Roact.Type and Roact.typeOf to expose a type checking API for external use

Checklist before submitting:

### DIFF
--- a/src/Type.lua
+++ b/src/Type.lua
@@ -15,9 +15,12 @@ local strict = require(script.Parent.strict)
 local Type = newproxy(true)
 
 local TypeInternal = {}
+local TypeNames = {}
 
 local function addType(name)
-	TypeInternal[name] = Symbol.named("Roact" .. name)
+	local symbol = Symbol.named("Roact" .. name)
+	TypeNames[symbol] = name
+	TypeInternal[name] = symbol
 end
 
 addType("Binding")
@@ -37,6 +40,14 @@ function TypeInternal.of(value)
 	return value[Type]
 end
 
+function TypeInternal.nameOf(type)
+	if typeof(type) ~= "userdata" then
+		return nil
+	end
+
+	return TypeNames[type]
+end
+
 getmetatable(Type).__index = TypeInternal
 
 getmetatable(Type).__tostring = function()
@@ -44,5 +55,6 @@ getmetatable(Type).__tostring = function()
 end
 
 strict(TypeInternal, "Type")
+strict(TypeNames, "TypeNames")
 
 return Type
--- a/src/TypeMirror.lua
+++ b/src/TypeMirror.lua
@@ -1,0 +1,48 @@
+local Type = require(script.Parent.Type)
+local Symbol = require(script.Parent.Symbol)
+local strict = require(script.Parent.strict)
+
+local ALLOWED_TYPES = {
+	Type.Binding,
+	Type.Element,
+	Type.HostChangeEvent,
+	Type.HostEvent,
+	Type.StatefulComponentClass,
+	Type.StatefulComponentInstance,
+	Type.VirtualTree
+}
+
+local MirroredType = newproxy(true)
+local MirroredTypeInternal = {}
+for _, type in ipairs(ALLOWED_TYPES) do
+	local name = Type.nameOf(type)
+	MirroredTypeInternal[name] = Symbol.named("Roact" .. name)
+end
+
+getmetatable(MirroredType).__index = MirroredTypeInternal
+getmetatable(MirroredType).__tostring = function()
+	return "RoactType"
+end
+
+strict(MirroredTypeInternal, "Type")
+
+local Mirror = newproxy(true)
+local MirrorInternal = {
+	Type = MirroredType,
+	typeOf = function(value)
+		local name = Type.nameOf(Type.of(value))
+		if not name then
+			return nil
+		end
+		return MirroredTypeInternal[name]
+	end,
+}
+
+getmetatable(Mirror).__index = MirrorInternal
+getmetatable(Mirror).__tostring = function()
+	return "TypeMirror"
+end
+
+strict(MirrorInternal, "TypeMirror")
+
+return Mirror
--- a/src/TypeMirror.lua
+++ b/src/TypeMirror.lua
@@ -1,3 +1,13 @@
+--[[
+	Mirrors a subset of values from Type.lua for external use, allowing
+	type checking on Roact objects without exposing internal Type symbols
+
+	TypeMirror: {
+		Type: Roact.Type,
+		typeof: function(value: table) -> Roact.Type | nil
+	}
+]]
+
 local Type = require(script.Parent.Type)
 local Symbol = require(script.Parent.Symbol)
 local strict = require(script.Parent.strict)
--- a/src/TypeMirror.spec.lua
+++ b/src/TypeMirror.spec.lua
@@ -1,0 +1,64 @@
+return function()
+	local Type = require(script.Parent.Type)
+	local Mirror = require(script.Parent.TypeMirror)
+	local allowedTypes = {
+		Type.Binding,
+		Type.Element,
+		Type.HostChangeEvent,
+		Type.HostEvent,
+		Type.StatefulComponentClass,
+		Type.StatefulComponentInstance,
+		Type.VirtualTree
+	}
+
+	describe("Type", function()
+		it("should return a mirror of an internal type", function()
+			local name = Type.nameOf(Type.Element)
+			local mirroredType = Mirror.Type[name]
+			expect(mirroredType).to.equal(Mirror.Type.Element)
+		end)
+
+		it("should not return the actual internal type", function()
+			local name = Type.nameOf(Type.Element)
+			local mirroredType = Mirror.Type[name]
+			expect(mirroredType).to.never.equal(Type.Element)
+		end)
+
+		it("should include all allowed types", function()
+			for _, type in ipairs(allowedTypes) do
+				local name = Type.nameOf(type)
+				local mirroredType = Mirror.Type[name]
+				expect(mirroredType).to.be.ok()
+			end
+		end)
+
+		it("should not include any other types", function()
+			local name = Type.nameOf(Type.VirtualNode)
+			local success = pcall(function()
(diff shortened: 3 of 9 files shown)


src/TypeMirror.lua
```diff
@@ -0,0 +1,58 @@
...
+	typeOf = function(value)
+		local name = Type.nameOf(Type.of(value))
+		if not name then
+			return nil
+		end
+		return MirroredTypeInternal[name]
+	end,
+}
+
+getmetatable(Mirror).__index = MirrorInternal
```
dbrooks-roblox: Do you want to disable __newindex?

MisterUncloaked: That happens by using strict() below

src/TypeMirror.lua
```diff
@@ -0,0 +1,58 @@
+--[[
+	Mirrors a subset of values from Type.lua for external use, allowing
+	type checking on Roact objects without exposing internal Type symbols
+
+	TypeMirror: {
+		Type: Roact.Type,
+		typeof: function(value: table) -> Roact.Type | nil
```
jeparlefrancais:
```suggestion
		typeOf: function(value: table) -> Roact.Type | nil
```

docs/advanced/inspecting-roact-objects.md
```diff
@@ -0,0 +1,71 @@
...
+end
+```
+
+## With Object Type Inspection
+
+With `Roact.typeOf`, we can be certain we have a Component class:
+```lua
+Header.validateProps = function()
+	local titleClass = props.titleClass
+	if Roact.typeOf(titleClass) == Roact.Type.StatefulComponentClass then
```
LPGhatguy: This is something I hadn't thought about before -- do we want people writing checks like this? Since a component can be a `StatefulComponentClass`, any `function`, or any `string`, it feels odd to specifically check for one of these options.

src/Type.lua
```diff
@@ -37,12 +40,21 @@ function TypeInternal.of(value)
 	return value[Type]
 end
 
+function TypeInternal.nameOf(type)
+	if typeof(type) ~= "userdata" then
+		return nil
+	end
```
LPGhatguy: If `TypeNames` is only indexed by userdata anyways, this check is redundant.

src/TypeMirror.lua
```diff
@@ -0,0 +1,58 @@
...
+}
+
+getmetatable(Mirror).__index = MirrorInternal
+getmetatable(Mirror).__tostring = function()
+	return "TypeMirror"
+end
+
+strict(MirrorInternal, "TypeMirror")
+
+return Mirror
```
LPGhatguy: Can we avoid the userdata->table dance for this module? We did that for `Type` since the userdata `Type` is also used as the key into the actual objects. Then we can apply `strict` to the table instead of needing to apply it to an inner table.

src/TypeMirror.spec.lua
```diff
@@ -0,0 +1,64 @@
...
+	local Mirror = require(script.Parent.TypeMirror)
+	local allowedTypes = {
+		Type.Binding,
+		Type.Element,
+		Type.HostChangeEvent,
+		Type.HostEvent,
+		Type.StatefulComponentClass,
+		Type.StatefulComponentInstance,
+		Type.VirtualTree
+	}
```
LPGhatguy: This is interesting redundancy. Do you think there's any risk that we'll forget to update this table, which would allow tests to pass erroneously?

docs/advanced/inspecting-roact-objects.md
jeparlefrancais: It's kind of like bringing ElementKind accessible at this point, but I am not sure if we want that or not.

MisterUncloaked: I think you're right. And even `ElementKind` has limited use from a developer's point of view since you have to have already created the element to use it. After talking to @dbrooks-roblox , I think we need to add Roact.isValidElementType in this PR - (basically [this](https://github.com/facebook/react/blob/master/packages/shared/isValidElementType.js)

src/TypeMirror.spec.lua
MisterUncloaked: That's a good point - TypeMirror should return a strict table with this list instead

src/isValidElementType.lua
```diff
@@ -0,0 +1,14 @@
+local Portal = require(script.Parent.Portal)
+local Type = require(script.Parent.Type)
+
+-- Returns true if the provided object can be used by Roact.createElement
+return function(value)
```
MisterUncloaked: I'll add more tests and docs on Monday, but here's isValidElementType for consumers who want to see if a value can be passed to Roact.createElement, which is useful for building APIs that can take any component type

LPGhatguy: Maybe that method should be called `isComponent`, since that's all an 'element type' is?

MisterUncloaked: I went with `isValidElementType` to match the API in [react-is](https://www.npmjs.com/package/react-is#usage) but I agree that API name is more clear.

MisterUncloaked:
This PR is ready for another pass - tests and docs are updated with Roact.isComponent (formerly Roact.isValidElementType)

src/TypeMirror.spec.lua
```diff
@@ -0,0 +1,56 @@
...
+		it("should throw if the value is not a valid type", function()
+			local typeOfCheck = function(value)
+				local _ = Mirror.typeOf(value)
+			end
+			expect(pcall(typeOfCheck, 1)).to.equal(false)
+			expect(pcall(typeOfCheck, true)).to.equal(false)
+			expect(pcall(typeOfCheck, "test")).to.equal(false)
+			expect(pcall(typeOfCheck, print)).to.equal(false)
+			expect(pcall(typeOfCheck, {})).to.equal(false)
+			expect(pcall(typeOfCheck, newproxy(true))).to.equal(false)
```
jeparlefrancais: Here I am not sure why you need to wrap your function call, I thought you had a function that returns a function that should throw
```lua
local function getTypeCheckCallback(value)
    return function()
        Mirror.typeOf(value)
    end
end
-- then the assert can be
expect(getgetTypeCheckCallback(1)).to.throw()
```
Or you could just not wrap your function and do
```lua
expect(pcall(Mirror.typeOf, 1)).to.equal(false)
```

But this also makes me think that with the new update of `testez` expect will throw if given multiple arguments (like pcall does when the function throws), so it's probably better to wrap it!

src/isComponent.spec.lua
```diff
@@ -0,0 +1,33 @@
...
+	-- There's no way to guarantee the return type for a function in Lua at the moment
+	itSKIP("should not return true for a function that returns an invalid type", function() end)
+
+	it("should return true for a string representing a host instance type", function()
+		local host = "Frame"
+		expect(isComponent(host)).to.equal(true)
+	end)
+
+	-- In the future, an exhaustive enum of all possible host instance types could enable this check
+	itSKIP("should not return true for a function that returns an invalid type", function() end)
```
jeparlefrancais: Oops you left a skip here and a little above too!

MisterUncloaked: See comment above this line - the skips are intentional because the checks can't easily be written today

jeparlefrancais: Oh, my bad! Both tests have the same name though

MisterUncloaked:
This PR should be ready to go at this point.

src/TypeMirror.lua
```diff
@@ -0,0 +1,59 @@
...
+		end
+		return MirroredType[name]
+	end,
+}
+
+setmetatable(Mirror, {
+	__tostring = function()
+		return "TypeMirror"
+	end
+})
```
LPGhatguy: Should we bother giving this table its own name? Users of Roact don't care that it's a mirror of our internals.

src/TypeMirror.spec.lua
```diff
@@ -0,0 +1,53 @@
...
+
+	describe("typeOf", function()
+		it("should throw if the value is not a valid type", function()
+			expect(pcall(Mirror.typeOf, 1)).to.equal(false)
+			expect(pcall(Mirror.typeOf, true)).to.equal(false)
+			expect(pcall(Mirror.typeOf, "test")).to.equal(false)
+			expect(pcall(Mirror.typeOf, print)).to.equal(false)
+			expect(pcall(Mirror.typeOf, {})).to.equal(false)
+			expect(pcall(Mirror.typeOf, newproxy(true))).to.equal(false)
+		end)
```
LPGhatguy: Should these throw? Looking at the code it doesn't seem like they should, and I don't think they should from a type perspective. We don't know if something is a Roact object until we call `typeOf` on it!

docs/advanced/type-validation.md
```diff
@@ -0,0 +1,81 @@
...
+	}, {
+		Title = Roact.isComponent(title) and Roact.createElement(title, {
+			-- Props for Title...
+		})
+	})
+end
+```
+
+!!! info
+	Because strings (hosts) and functions are valid component types, `Roact.isComponent` is less safe than `Roact.typeOf`. If safety is paramount, consider only allowing component classes, and checking that the `typeOf` the prop is `Roact.Type.StatefulComponentClass`.
```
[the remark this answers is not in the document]
MisterUncloaked: Good point. We can probably omit this piece entirely. The thought was that `isComponent` can return true for invalid host strings and/or functions, and you can avoid that problem by only accepting `Roact.Type.StatefulComponentClass` but maybe that's not practical.

src/TypeMirror.spec.lua
MisterUncloaked: It throws because none of these are valid keys when getting the `nameOf` the real type, and it seemed reasonable to throw if you're not passing in a plausible symbol (userdata), but I think not throwing and returning nil sounds good too. Updated.

MisterUncloaked:
@LPGhatguy anything else you'd like to see here before we get this merged?

PR #236, opened 2019-09-18
Add scripts to generate release models

LPGhatguy:
This adds a script or two to the `bin` folder to help us generate good release models.

We can use this in the future as part of our CI flow, but in the mean time, it should serve as light automation to generate builds at any point that can be dragged into Roblox Studio.

We should wait for `rbxm` support to be ironed out in Remodel before merging this, and switch everything over to using the binary model format instead.

### DIFF
--- a/bin/build-release.sh
+++ b/bin/build-release.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+# Generates a release build of Roact with tests stripped.
+#
+# Usage from repo root:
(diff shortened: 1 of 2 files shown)


LPGhatguy:
We should figure out a good way to get these tools onto our CI machines (like [Foreman](https://github.com/rojo-rbx/foreman)) so that we can run them automatically and also pin versions.

PR #242, opened 2019-10-09
Pass props and state through render

### LINKED ISSUE #199: Make StatefulComponent:render accept props and state as arguments
This is something that Preact does, but React doesn't. It helps reduce the number of times we need to refer to `self.props` and `self.state` and replaces them with just `props` and `state`.

Kampfkarren:
Closes #199

Checklist before submitting:

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -400,20 +400,30 @@ end
 
 ### render
 ```
-render() -> Element | nil
+render(props, state) -> Element | nil
 ```
 
 `render` describes what a component should display at the current instant in time.
+
+```lua
+function MyComponent:render(props, state)
+	-- props == self.props
+	-- state == self.state
+	return Roact.createElement("TextLabel", {
+		Text = "Hello, " .. props.name .. "!"
+	})
+end
+```
 
 !!! info
 	Roact assumes that `render` act likes a pure function: the result of `render` must depend only on `props` and `state`, and it must not have side-effects.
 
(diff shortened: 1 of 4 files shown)


docs/api-reference.md
```diff
@@ -400,20 +400,30 @@ end
...
-render() -> Element | nil
+render(props, state) -> Element | nil
 ```
 
 `render` describes what a component should display at the current instant in time.
 
+```lua
+function MyComponent:render(props, state)
+	-- props == self.props
+	-- state == self.state
```
ZoteTheMighty: We might want to add a little bit more explanation of what we're trying to convey here. Maybe a little blurb that says that props and state can be accessed either as arguments or as properties on self?

Kampfkarren: I can do that, I was just copying Preact's documentation.

ZoteTheMighty:
This still looks good to me, but I did some digging into why React doesn't do this, and here's what I found: https://github.com/reactjs/rfcs/pull/4#issuecomment-412637588

It's a pretty important point to make! Is it something we're worried about for Roact? Is it something we think we'll need to worry about? I may be missing something, but I think the closing-over-this problem they're talking about applies pretty equivalently to Lua as well as JS

ZoteTheMighty:
One more thing to note as a followup: we do also have the same problem already with initial props passed into `init`, so we should keep that in mind as well when we consider this.

PR #246, opened 2020-01-09
Add Context API

### LINKED ISSUE #4: New Context API

psewell:
Adds createContext, provide, and consume APIs to Roact.
Components that consume Contexts automatically update when the Context updates, wherever they are in the tree.

Closes #4 .

- Adds createContext function, which is used to make a new Context
- createContext returns a Provider and Consumer.

```lua

local function UsingValue()
		render = function(theme)
				Text = theme,
			})
		end,
	})
end

local function Root()
		value = "overridden value",
	}, {
	})
end
```

Checklist before submitting:


src/createContext.lua
```diff
@@ -0,0 +1,92 @@
...
+
+	function Provider:init(props)
+		self.binding, self.updateValue = Binding.create(props.value)
+
+		local key = context.key
+		self._context[key] = self.binding
+	end
+
+	function Provider:didUpdate()
+		self.updateValue(self.props.value)
```
LPGhatguy: Let's check that `prevProps.value ~= self.props.value` before firing this method. This will let us avoid firing updates when we don't need to.

src/createContext.lua
```diff
@@ -0,0 +1,92 @@
...
+	end
+
+	function Provider:render()
+		return oneChild(self.props[Children])
```
LPGhatguy: We should use a fragment here instead! We added them in 1.0 to improve cases like this, and will let providers have multiple children instead of just one.

src/createContext.lua
```diff
@@ -0,0 +1,92 @@
+local Symbol = require(script.Parent.Symbol)
+local Binding = require(script.Parent.Binding)
+local oneChild = require(script.Parent.oneChild)
+local Children = require(script.Parent.PropMarkers.Children)
+local PureComponent = require(script.Parent.PureComponent)
+
+local function createProvider(context)
+	local Provider = PureComponent:extend("Provider")
```
LPGhatguy: Let's use a regular component here. Because this component is always going to have children, the `shouldUpdate` check provided by `PureComponent` is almost always going to fail.

In general, we should prefer function components or `Component` by default, and fall back to `PureComponent` with profiling if we have performance issues.

src/createContext.lua
```diff
@@ -0,0 +1,92 @@
...
+		else
+			-- Fall back to the default value if no Provider exists
+			self.state = {
+				value = context.defaultValue,
+			}
+		end
+	end
+
+	function Consumer:render()
+		assert(type(self.props.render) == "function", "Consumer expects a `render` function")
```
LPGhatguy: We should be able to use Roact's `validateProps` feature here instead of asserting in `render`!

src/createContext.lua
```diff
@@ -0,0 +1,92 @@
...
+		return self.props.render(self.state.value)
+	end
+
+	function Consumer:willUnmount()
+		if self.disconnect then
```
LPGhatguy: Can we be a little bit more explicit with this check, as in

```suggestion
		if self.disconnect ~= nil then
```

It's unlikely that `disconnect` will be `false`, but we've been trying to encourage that explicitness when the exact type involved is unclear.

src/createContext.lua
```diff
@@ -0,0 +1,92 @@
...
+	local self = {
+		defaultValue = defaultValue,
+		key = Symbol.named("ContextKey"),
+	}
+	setmetatable(self, Context)
+	return self
+end
+
+function Context:__tostring()
+	return tostring(self.defaultValue)
```
LPGhatguy: This is kind of a strange `__tostring`!

Can we return a fixed string here like `RoactContext`?

src/createContext.spec.lua
```diff
@@ -0,0 +1,168 @@
...
+		expect(context.Provider).to.be.ok()
+		expect(context.Consumer).to.be.ok()
+	end)
+
+	describe("Provider", function()
+		it("should render its child component", function()
+			local didRender = false
+			local context = createContext("Test")
+
+			local Listener = PureComponent:extend("Listener")
```
LPGhatguy: Instead of a `PureComponent`, this should just be a function component for simplicity.

src/createContext.spec.lua
```diff
@@ -0,0 +1,168 @@
...
+
+			function Listener:render()
+				return nil
+			end
+
+			local element = createElement(context.Consumer)
+
+			expect(function()
+				local tree = noopReconciler.mountVirtualTree(element, nil, "Provide Tree")
+				noopReconciler.unmountVirtualTree(tree)
```
LPGhatguy: We don't need to call `unmountVirtualTree`. We should limit the scope of what's inside this `expect` block so that we don't erroneously pass this test when something else throws.

src/createContext.spec.lua
```diff
@@ -0,0 +1,168 @@
...
+				render = function(value)
+					foundValue = value
+				end,
+			})
+
+			local tree = noopReconciler.mountVirtualTree(element, nil, "Provide Tree")
+			noopReconciler.unmountVirtualTree(tree)
+
+			expect(foundValue).to.equal("Test")
+		end)
```
LPGhatguy: I think it would be more robust to use Roact's internal `createSpy` method to create a spy function to use for the render function. This lets us assert how many times it was called, as well as _all_ of the arguments passed.

Notably, this lets us catch errors like:
1. We invoke the function twice and the first one is incorrect
2. We pass extra arguments to the function

src/createContext.spec.lua
```diff
@@ -0,0 +1,168 @@
...
+				value = "ThirdTest",
+			}, {
+				Listener = createElement(Listener),
+			}))
+
+			expect(renderCount).to.equal(2)
+			expect(foundValue).to.equal("ThirdTest")
+
+			noopReconciler.unmountVirtualTree(tree)
+		end)
```
LPGhatguy: This test can also be rewritten in terms of Roact's spies, which might be more readable.

PR #247, opened 2020-01-10
New internal Context API

### LINKED ISSUE #4: New Context API

ZoteTheMighty:
Works towards #4. This is a prerequisite for the new _public_ Context API proposed in #246.

This change introduced two new methods to the `Component` API:

* `Component:__addContext(key, value)` - Adds a new entry to the context table. This will be available to this component and all of its children, and is intended to be used in a Provider component.
* `Component:__getContext(key)` - Retrieves the context value for the given key. This intended to be used by a Consumer component.

This interface allows us to avoid defensive copying, as even internal Component APIs do not get direct access to the context table. The legacy context API is still present and usable, and will not be affected by these changes.

Checklist before submitting:
* ~~Added entry to `CHANGELOG.md`~~ (this should come with the changes in #246)
* [x] Added/updated documentation (documentation here consists of code comments, public docs will be part of #246)

### DIFF
--- a/src/Component.spec/context.spec.lua
+++ b/src/Component.spec/context.spec.lua
@@ -1,0 +1,297 @@
+return function()
+	local assertDeepEqual = require(script.Parent.Parent.assertDeepEqual)
+	local createElement = require(script.Parent.Parent.createElement)
+	local createReconciler = require(script.Parent.Parent.createReconciler)
+	local NoopRenderer = require(script.Parent.Parent.NoopRenderer)
+	local oneChild = require(script.Parent.Parent.oneChild)
+
+	local Component = require(script.Parent.Parent.Component)
+
+	local noopReconciler = createReconciler(NoopRenderer)
+
+	it("should be provided as an internal api on Component", function()
+		local Provider = Component:extend("Provider")
+
+		function Provider:init()
+			self:__addContext("foo", "bar")
+		end
+
+		function Provider:render()
+		end
+
+		local element = createElement(Provider)
+		local hostParent = nil
+		local hostKey = "Provider"
+		local node = noopReconciler.mountVirtualNode(element, hostParent, hostKey)
+
+		local expectedContext = {
+			foo = "bar",
+		}
+
+		assertDeepEqual(node.context, expectedContext)
+	end)
+
+	it("should be inherited from parent stateful nodes", function()
+		local Consumer = Component:extend("Consumer")
+
+		local capturedContext
+		function Consumer:init()
+			capturedContext = {
+				hello = self:__getContext("hello"),
+				value = self:__getContext("value"),
+			}
+		end
+
+		function Consumer:render()
+		end
+
+		local Parent = Component:extend("Parent")
+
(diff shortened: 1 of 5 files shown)


src/Component.spec/context.spec.lua
```diff
@@ -92,12 +99,85 @@ return function()
...
+		local expectedContext = {
+			parent = "I'm here!",
+		}
+
+		assertDeepEqual(parentNode.context, expectedContext)
+
+		local childNode = oneChild(parentNode.children)
+
+		-- Parent and child should have the same context table
+		expect(parentNode.context, childNode.context)
```
LPGhatguy: Do you mean to use `assert(a == b)` here?

src/createReconciler.lua
```diff
@@ -247,10 +258,14 @@ local function createReconciler(renderer)
...
 		Constructs a new virtual node but not does mount it.
 	]]
-	local function createVirtualNode(element, hostParent, hostKey, context)
+	local function createVirtualNode(element, hostParent, hostKey, context, legacyContext)
 		if config.internalTypeChecks then
 			internalAssert(renderer.isHostObject(hostParent) or hostParent == nil, "Expected arg #2 to be a host object")
 			internalAssert(typeof(context) == "table" or context == nil, "Expected arg #4 to be of type table or nil")
+			internalAssert(
+				typeof(legacyContext) == "table" or legacyContext == nil,
+				"Expected arg #4 to be of type table or nil"
```
LPGhatguy:
```suggestion
				"Expected arg #5 to be of type table or nil"
```

Typo, this is arg 5.

src/createReconciler.lua
```diff
@@ -304,10 +330,13 @@ local function createReconciler(renderer)
...
 		the tree.
 	]]
-	function mountVirtualNode(element, hostParent, hostKey, context)
+	function mountVirtualNode(element, hostParent, hostKey, context, legacyContext)
 		if config.internalTypeChecks then
 			internalAssert(renderer.isHostObject(hostParent) or hostParent == nil, "Expected arg #2 to be a host object")
-			internalAssert(typeof(context) == "table" or context == nil, "Expected arg #4 to be of type table or nil")
+			internalAssert(
+				typeof(legacyContext) == "table" or legacyContext == nil,
+				"Expected arg #4 to be of type table or nil"
```
LPGhatguy:
```suggestion
				"Expected arg #5 to be of type table or nil"
```

Same copy-paste typo here.

src/Component.spec/context.spec.lua
ZoteTheMighty: I sure did! Fortunately, it still passes.

PR #250, opened 2020-01-14
Test change for trying GitHub Actions from a fork

LPGhatguy:
This is trying out #249 more authentically than it was able to on its own!


LPGhatguy:
Coveralls didn't happen... at all! Neither Travis nor GitHub Actions got Coveralls to comment on this issue? :/

PR #253, opened 2020-01-28
Document new context API and fix some docs mishaps

### LINKED ISSUE #207: Dividers on docs are not solid lines (and look ugly)
This is particularly bad with DPI scaling.

image

### LINKED ISSUE #241: Update docs so that it is clear validateProps does not take (self) as an an argument
Update docs so that it is clear validateProps does not take (self) as an an argument. This is an easy mistake to make, since the other methods require self.

LPGhatguy:
Fixes #207. Fixes #241.

This PR documents the new context API (#4) that was recently merged while cleaning up some of the little problems about our docs.

* Changed `<hr>` tags to be solid lines instead of dashed ones (#207)
* Added extra `static` notice to `validateProps` (#241)
* Removed custom CSS in favor of using [Admonition](https://squidfunk.github.io/mkdocs-material/extensions/admonition/)
* Added baseline documentation for the new context feature

### DIFF
--- a/docs/advanced/context.md
+++ b/docs/advanced/context.md
@@ -5,7 +5,7 @@
 
 [TOC]
 
-Roact supports a feature known as context that enables pass values down the tree without having to pass them through props. Roact's Context API is based on [React's Context API](https://reactjs.org/docs/context.html).
+Roact supports a feature called context which enables passing values down the tree without having to pass them through props. Roact's Context API is based on [React's Context API](https://reactjs.org/docs/context.html).
 
 Context is commonly used to implement features like dependency injection, dynamic theming, and scoped state storage.
 
--- a/docs/advanced/context.md
+++ b/docs/advanced/context.md
@@ -5,17 +5,91 @@
 
 [TOC]
 
-Roact supports a feature known as context that helps pass values down the tree without having to pass them through props.
+Roact supports a feature known as context that enables pass values down the tree without having to pass them through props. Roact's Context API is based on [React's Context API](https://reactjs.org/docs/context.html).
+
+Context is commonly used to implement features like dependency injection, dynamic theming, and scoped state storage.
 
 ## Basic Usage
-Context is defined by creating a pair of components known as the _Provider_ and the _Consumer_. Roact does this for you with the `Roact.createContext()` API:
-
 ```lua
-local MyValueContext = Roact.createContext()
+local ThemeContext = Roact.createContext(defaultValue)
 ```
 
-## Example: Theming
+Context objects contain two components, `Consumer` and `Provider`.
 
+The `Consumer` component accepts a `render` function as its only prop, which is used to render its children. It's passed one argument, which is the context value from the nearest matching `Provider` ancestor.
+
+If there is no `Provider` ancestor, then `defaultValue` will be passed instead.
+
+```lua
+local function ThemedButton(props)
+	return Roact.createElement(ThemeContext.Consumer, {
+		render = function(theme)
+			return Roact.createElement("TextButton", {
+				Size = UDim2.new(0, 100, 0, 100),
+				Text = "Click Me!",
+				TextColor3 = theme.foreground,
+				BackgroundColor3 = theme.background,
+			})
+		end
+	})
+end
+```
+
+The `Provider` component accepts a `value` prop as well as children. Any of its descendants will have access to the value provided to it by using the `Consumer` component like above.
+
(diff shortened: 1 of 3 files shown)


docs/advanced/context.md
```diff
@@ -1,5 +1,95 @@
...
+	This API is not yet available in a stable Roact release.
 
-!!! info
-	This section is incomplete. It's possible that the context API will change before the existing API is ever documented.
\ No newline at end of file
+	It may be available from a recent pre-release or Roact's master branch.
+
+[TOC]
+
+Roact supports a feature known as context that enables pass values down the tree without having to pass them through props. Roact's Context API is based on [React's Context API](https://reactjs.org/docs/context.html).
```
ZoteTheMighty: This is a nitpick, but wording is a bit odd.
```suggestion
Roact supports a feature called context which enables passing values down the tree without having to pass them through props. Roact's Context API is based on [React's Context API](https://reactjs.org/docs/context.html).
```

docs/advanced/context.md
```diff
@@ -1,5 +1,95 @@
...
+end
+
+function ThemeController:render()
+	return Roact.createElement(ThemeContext.Provider, {
+		value = self.state.theme,
+	}, self.props[Roact.Children])
+end
+```
+
+## Legacy Context
```
ZoteTheMighty: Are we worried folks will be more willing to add new uses of this now that it's documented in some detail? Calling something "deprecated" hasn't often stopped people before o.O

docs/api-reference.md
```diff
@@ -176,6 +179,39 @@ Creates a new reference object that can be used with [Roact.Ref](#roactref).
...
+
+`Provider` and `Consumer` are both Roact components.
+
+#### `Provider`
+`Provider` accepts the following props:
+
+* `value`: The value to put into the tree for this context value.
+	* If the `Provider` is updated with a new `value`, any matching `Consumer` components will be re-rendered with the new value.
+* `[Children]`: Any number of children to render underneath this provider.
+	* Descendants of this component can receiver the provided context value by using `Consumer`.
```
ZoteTheMighty:
```suggestion
	* Descendants of this component can receive the provided context value by using `Consumer`.
```

docs/advanced/context.md
```diff
@@ -1,5 +1,95 @@
...
+```lua
+local Consumer = Roact.Component:extend("FooConsumer")
+
+function Consumer:init()
+	self.foo = self._context[FooKey]
+end
+
+function Consumer:render()
+	return Roact.createElement("TextLabel", {
+		Text = "Foo: " .. self.foo.value,
```
ZoteTheMighty: Should maybe reiterate that this will not update immediately when context updates, but will be up-to-date on the next render?

docs/api-reference.md
```diff
@@ -176,6 +179,39 @@ Creates a new reference object that can be used with [Roact.Ref](#roactref).
...
+
+	It may be available from a recent pre-release or Roact's master branch.
+
+```
+Roact.createContext(defaultValue: any) -> { Provider, Consumer }
+```
+
+Creates a new context provider and consumer. For a usage guide, see [Advanced Concepts: Context](/advanced/context).
+
+`defaultValue` is given to consumers if they have no `Provider` ancestors. It is up to consumers of Roact's context API to turn this case into an error if it is an invalid state.
```
ZoteTheMighty: Seems like the overload of the word "consumer" is a bit confusing. Maybe this change makes sense?
```suggestion
`defaultValue` is given to consumers if they have no `Provider` ancestors. It is up to users of Roact's context API to turn this case into an error if it is an invalid state.
```

docs/api-reference.md
```diff
@@ -176,6 +179,39 @@ Creates a new reference object that can be used with [Roact.Ref](#roactref).
...
 
+### Roact.createContext
+
+!!! danger "Unreleased API"
+	This API is not yet available in a stable Roact release.
+
+	It may be available from a recent pre-release or Roact's master branch.
+
+```
+Roact.createContext(defaultValue: any) -> { Provider, Consumer }
```
ZoteTheMighty: Is the return type here clear enough? There's more detail in the advanced section, but I'm a bit concerned someone might think the return is an array instead of a table. Not sure how to easily reflect that with the syntax we've been using for type definitions.

docs/advanced/context.md
LPGhatguy: Maybe we should add more `!!! danger` blocks!

docs/api-reference.md
LPGhatguy: 😭

MisterUncloaked:
lgtm

PR #260, opened 2020-03-18
Fix prop-context update tearing

### LINKED ISSUE #259: New context API has delayed update behavior
If I update a context value in the same update as adding a consumer of that value, the consumer will receive the old update. I'm still working on a better standalone repro, but basically:

```lua
local tree = Roact.mount(
	Roact.createElement(Foo.Provider, {
		value = 1,
	})
)

Roact.update(
	tree,
	Roact.createElement(Foo.Provider, {
		value = 2,
	}, {
		Consumer = Roact.createElement(Foo.Consumer, {
			render = function(value)
				print("value", value)
			end
		}),
	})
)
```

This code should print `value 2`, but I believe (once I pull a repro out) that it would instead print `value 1`.

This manifested itself as a bug in a game that I'm working on.

LPGhatguy:
Closes #259.

This PR tries to fix the update tearing behavior I discovered and reported in #259 based on a solution that my team has been discussing. I introduced a fairly small and robust test that is currently failing to indicate the problem with our current state and the quick-fix solutions to this problem.

Context values are no longer based on bindings and now instead use a small cell that contains a value and an update signal.

Providers now update context values in two passes:
1. In `willUpdate`, the provider updates the value of the context entry but do not fire any signals. This should cause downstream consumers to be able to immediately read the new value on their next render. Their next render will occur immediately unless an intermediate component implements `shouldUpdate`.
2. In `didUpdate`, the provider fires the signal portion of the context entry. This will notify any consumers.

Consumers are now a little dumber and simpler. They no longer track the current context value in state, but instead track the last value they were updated with. When an update signal is received from their corresponding provider's `didUpdate` hook, consumers now compare the new value with the value they were last updated with, stored in their `didUpdate` hook. If these values differ, we can assume that there was a component between the consumer and provider that returned `false` to `shouldUpdate`, and trigger an update.

I dramatically increased the documentation around affected code to make these kinds of issues more clear to us in the future.

## TODO
* [x] Failing test and bug explanation in source code
* [x] Actually fix behavior

Checklist before submitting:
* ~Added entry to `CHANGELOG.md`~
    * We didn't make a release with the broken context API, so this should roll up under just introducing this feature.
* ~Added/updated documentation~
    * No user-visible changes because this feature has not been released yet.

### DIFF
--- a/src/createContext.lua
+++ b/src/createContext.lua
@@ -102,7 +102,9 @@ local function createConsumer(context)
 			--
 			-- To avoid sending a redundant update, we compare the new value
 			-- with the last value that we updated with (set in didUpdate) and
-			-- only update if they differ.
+			-- only update if they differ. This may happen when an update from a
+			-- provider was blocked by an intermediate component that returned
+			-- false from shouldUpdate.
 			self.disconnect = self.contextEntry.onUpdate:subscribe(function(newValue)
 				if newValue ~= self.lastValue then
 					-- Trigger a dummy state update.
--- a/src/createContext.lua
+++ b/src/createContext.lua
@@ -34,7 +34,7 @@ local function createProvider(context)
 	end
 
 	function Provider:didUpdate(prevProps)
-		-- If the provided value change, after we've updated every reachable
+		-- If the provided value changed, after we've updated every reachable
 		-- component, fire a signal to update the rest.
 		--
 		-- This signal will notify all context consumers. It's expected that
--- a/src/createContext.lua
+++ b/src/createContext.lua
@@ -1,22 +1,50 @@
 local Symbol = require(script.Parent.Symbol)
-local Binding = require(script.Parent.Binding)
 local createFragment = require(script.Parent.createFragment)
+local createSignal = require(script.Parent.createSignal)
 local Children = require(script.Parent.PropMarkers.Children)
 local Component = require(script.Parent.Component)
+
+--[[
+	Construct the value that is assigned to Roact's context storage.
+]]
+local function createContextEntry(currentValue)
+	return {
+		value = currentValue,
+		onUpdate = createSignal(),
+	}
+end
 
 local function createProvider(context)
 	local Provider = Component:extend("Provider")
 
 	function Provider:init(props)
-		self.binding, self.updateValue = Binding.create(props.value)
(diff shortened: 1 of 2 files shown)


src/createContext.lua
```diff
@@ -61,8 +66,50 @@ local function createConsumer(context)
...
+	function Consumer:didMount()
+		if self.contextEntry ~= nil then
+			-- When onUpdate is fired, a new value has been made available in
+			-- this context entry, but we may have already updated in the same
+			-- update cycle.
+			--
+			-- To avoid sending a redundant update, we compare the new value
+			-- with the last value that we updated with (set in didUpdate) and
+			-- only update if they differ.
+			self.disconnect = self.contextEntry.onUpdate:subscribe(function(newValue)
```
MisterUncloaked: When can this happen?

LPGhatguy: Updated the comment to make this more clear.

MisterUncloaked: Cool. Is it possible for our tests to cover this case?

src/createContext.lua
```diff
@@ -1,22 +1,50 @@
...
+		-- During this update, any components that are reachable will receive
+		-- this updated value at the same time as any props and state updates
+		-- that are being applied.
+		if nextProps.value ~= self.props.value then
+			self.contextEntry.value = nextProps.value
+		end
 	end
 
 	function Provider:didUpdate(prevProps)
+		-- If the provided value change, after we've updated every reachable
```
ZoteTheMighty:
```suggestion
		-- If the provided value changed, after we've updated every reachable
```

ZoteTheMighty:
This looks good! I think it solves it pretty nicely. I had a nagging feeling that somehow the Consumer should be expressing its render deduplication via `shouldUpdate`, but I'm not sure if that would work any better than this as it is.

src/createContext.lua
LPGhatguy: Good idea!

LPGhatguy: Just added a test for this. I don't think it's directly related to this change, but is still a great test to have.

MisterUncloaked:
This looks good to me. I like this solution a lot.

LPGhatguy:
Woops, forgot to mark as ready to review.

PR #261, opened 2020-03-25
Remove Travis-CI

LPGhatguy:
This change completes the transition to GitHub Actions, which was introduced for Roact CI in #249. I think we've established that it should work how we need it to, and this paves the way for us to adopt tools like TestEZ CLI.


LPGhatguy:
It looks like the Coveralls report succeeded, but it didn't get posted back to this PR: https://coveralls.io/builds/29613181

LPGhatguy:
Dang. Switched to the official Coveralls GitHub Action, which I'm happier with config-wise, but still has problems.

https://github.com/coverallsapp/github-action/issues/7

LPGhatguy:
👋 bye travis

PR #275, opened 2020-06-11
Add component name to property validation error message

ConorGriffin37:
Add componentName to property validation error message. This would be helpful in cases where element tracing is not turned on or provides an unhelpful error message.

Checklist before submitting:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased Changes
 
-* Added component name to property validation error message
+* Added component name to property validation error message ([#275](https://github.com/Roblox/roact/pull/275))
 
 ## [1.3.0](https://github.com/Roblox/roact/releases/tag/v1.3.0) (May 5th, 2020)
 * Added Contexts, which enables easy handling of items that are provided and consumed throughout the tree.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased Changes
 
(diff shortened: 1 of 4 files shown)


src/Component.spec/validateProps.spec.lua
```diff
@@ -188,7 +188,8 @@ return function()
 			end)
 
 			expect(success).to.equal(false)
-			expect(error:find("MyComponent")).to.be.ok()
+			local startIndex = error:find("MyComponent")
+			expect(startIndex).to.be.ok()
```
ZoteTheMighty, line 192: Oh, good catch here!

ConorGriffin37:
@cliffchapmanrbx Looks like this CLA check doesn't work with forks, not sure if we need to change it?

I'll just add myself to the whitelist since I work at Roblox I presume this doesn't apply to me.

cliffchapmanrbx:
Minor deployment bugs on the bot that I fixed late yesterday, it should work normally now. As it says at http://go/clabot FTEs should not sign the CLA, they should be whitelisted.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased Changes
 
+* Added component name to property validation error message
```
LPGhatguy, line 5:
```suggestion
* Added component name to property validation error message ([#275](https://github.com/Roblox/roact/pull/275))
```
Linking back to the PR containing a change is something we forget to do a lot, but is super useful when looking back over these!

PR #282, opened 2020-07-08
Disconnect events during unmount

### LINKED ISSUE #235: Disconnect event listeners on unmount
We don't currently disconnect events before we start unmounting host components.

This means that you can get event callbacks during unmounting, which means that refs are generally `nil`, you can't call `setState`, and generally things are in weird states.

We should update `RobloxRenderer.unmountHostNode` to iterate through all `Roact.Event` and `Roact.Change` props and disconnect them.

There may be cases in existing Roact projects where layout events are firing as children are being destroyed, and we might be doing a significant number of extraneous size changes.

jeparlefrancais:
Closes #235.

Change RobloxRenderer to disconnect all events during unmount, by adding a method `disconnectAll` to the SingleEventManager.

The test I added to RobloxRenderer is currently skipped, because lemur does not have the `ChildRemoved` event implemented. However, the test passes when testing in studio. Let me know if I should go implement lemur or if you have any other ideas about how I could test this.

Checklist before submitting:

### DIFF
--- a/src/RobloxRenderer.spec.lua
+++ b/src/RobloxRenderer.spec.lua
@@ -11,6 +11,7 @@ return function()
 	local GlobalConfig = require(script.Parent.GlobalConfig)
 	local Portal = require(script.Parent.Portal)
 	local Ref = require(script.Parent.PropMarkers.Ref)
+	local Event = require(script.Parent.PropMarkers.Event)
 
 	local RobloxRenderer = require(script.Parent.RobloxRenderer)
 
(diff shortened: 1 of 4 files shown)


src/RobloxRenderer.spec.lua
```diff
@@ -541,6 +542,25 @@ return function()
 			expect(spyRef.callCount).to.equal(2)
 			spyRef:assertCalledWith(nil)
 		end)
+
+		itSKIP("should not process events when unmounting", function()
```
matthargett, line 546: Should this still be skipped? If it's not currently passing, should it get to passing before being merged?

jeparlefrancais: I am not sure what I should do with this, because the test passes when it's running in studio, but not using Lemur, because it does not simulate the ChildRemoved event. So I don't know if I should implement the feature in Lemur or if there is another way to test this!

matthargett: @ZoteTheMighty how should we handle this?

PR #288, opened 2020-11-23
Experimental warning to help find connections that are not disconnected

DarraghGriffin:
This PR is quite speculative, a similar method to this helped track down a memory leak which I was investigating in a Roact project recently but I am not sure if something so specific can be justified as a Roact debugging configuration. I'd like to put up this PR though to see if anyone has any other ideas about how we could catch simple memory leaks like this within Roact as I think this is the most common type of memory leak in many projects and it seems like something we could help Roact consumers catch.

This PR introduces a new configuration option which aims to help Roact users more easily find components which are not properly disconnecting connections when unmounted and are potentially causing memory leaks because of this reason.

A full traversal of the component is performed to find all currently active connections but some keys are excluded to reduce false positives.

If a connection is referenced in a component but it's closure does not close over self as an upvalue then it is not a concern that the component still has a reference after being unmounted.
This commonly occurs through references to component context or the store. Ideally we would be able to inspect the closure associated with each active connection and warn only if the closure uses self as an upvalue.

Internal Roact connections are also ignored for this check, so that connections within EventManager do not cause this warning to be emitted.


DarraghGriffin:
I have read the CLA Document and I hereby sign the CLA

PR #294, opened 2021-01-14
Follow style guide + Luau performance guide

howmanysmall:
I've made the following changes to the codebase, as seen in the changelog.

* Fixed several linting problems.
* Made a few small performance changes (reducing indexing, checking optional values being nil before type checking them, using `string.x` instead of `:x`).
* Made the code better follow the [Roblox Lua Style Guide](https://roblox.github.io/lua-style-guide/).
* Changed some of the examples to also follow the Lua style guide better.

Checklist before submitting:
* [X] Added entry to `CHANGELOG.md`


examples/changed-signal/init.lua
```diff
@@ -40,9 +42,9 @@ return function()
 	local TextReverser = Roact.Component:extend("TextReverser")
 
 	function TextReverser:init()
-		self.state = {
```
matthargett, line 43: we may be changing our guidance on this in the future, so leave this one for now.

matthargett (commented):
Thanks for the contribution! A few minor requests, and then we can review again.

PR #302, opened 2021-05-07
Try bumping coveralls and checkout actions

ZoteTheMighty:
Trying to see if we can fix the coveralls reporting issues and unblock some of the outstanding PRs.


ZoteTheMighty:
Seems like it works!

matthargett (approved):

PR #307, opened 2021-05-25
Forward ref

ZoteTheMighty:
This closes a gap in compatibility with upstream React. While the current Roact implementation doesn't support refs assigned to non-host components, it also doesn't provide a way for non-host components to forward refs idiomatically as described here: https://reactjs.org/docs/forwarding-refs.html

This change introduces an upstream-compatible forwardRef API and loosely adapts some of the upstream tests as well.

Checklist before submitting:
* [x] Add a test to validate that the ref isn't also provided as a member of props


oltrep (approved):
LGTM

src/forwardRef.spec.lua
```diff
@@ -0,0 +1,341 @@
+-- Tests loosely adapted from those found at:
+-- * https://github.com/facebook/react/blob/v17.0.1/packages/react/src/__tests__/forwardRef-test.js
```
matthargett, line 2: put the LICENSE reference. and copyright information here as well

matthargett (approved):
definitely need to include the copyright and license info at the top of the derives test file. other than that, LGTM.

PR #315, opened 2021-08-05
Further update children re-rentrancy problems. 

ConorGriffin37:
See the previous pull request here:
https://github.com/Roblox/roact/pull/301

Originally I had thought that this issue was specific to events but it can actually also be caused using callbacks. We recently ran into this issue again and the original fix for this did not fix the issue because it was caused by callbacks and not events.

I'm still not fully sure what the best solution to this issue is, but I have two options that should work:
1. The original solution in the first PR. ( https://github.com/Roblox/roact/pull/301/commits/1ce1ea68a01664d972940145b3fdd7065e1af491 )
2. We could add a pendingProps mechanism similar to the pendingState mechanism and stop updates based on props changing during certain parts of the components lifecycle.

Checklist before submitting:

### DIFF
--- a/src/RobloxRenderer.spec.lua
+++ b/src/RobloxRenderer.spec.lua
@@ -1025,5 +1025,96 @@ return function()
 				reconciler.unmountVirtualNode(instance)
 			end)
 		end)
+
+		it("should not allow re-entrancy in updateChildren even with callbacks", function()
+			local configValues = {
+				tempFixUpdateChildrenReEntrancy = true,
+			}
+
+			GlobalConfig.scoped(configValues, function()
+				local LowestComponent = Component:extend("LowestComponent")
+
+				function LowestComponent:render()
+					return createElement("Frame")
+				end
+
+				function LowestComponent:didMount()
+					self.props.onDidMountCallback()
+				end
+
+				local ChildComponent = Component:extend("ChildComponent")
+
+				function ChildComponent:init()
+					self:setState({
+						firstTime = true
+					})
+				end
+
+				local childCoroutine
+
+				function ChildComponent:render()
+					if self.state.firstTime then
+						return createElement("Frame")
+					end
+
+					return createElement(LowestComponent, {
+						onDidMountCallback = self.props.onDidMountCallback
+					})
+				end
+
+				function ChildComponent:didMount()
+					childCoroutine = coroutine.create(function()
+						self:setState({
+							firstTime = false
+						})
+					end)
+				end
+
+				local ParentComponent = Component:extend("ParentComponent")
+
+				function ParentComponent:init()
+					self:setState({
+						count = 1
+					})
+
+					self.onDidMountCallback = function()
+						if self.state.count < 5 then
+							self:setState({
+								count = self.state.count + 1,
+							})
+						end
+					end
+				end
+
+				function ParentComponent:render()
+					return createElement("Frame", {
+
+					}, {
+						ChildComponent = createElement(ChildComponent, {
+							count = self.state.count,
+							onDidMountCallback = self.onDidMountCallback,
+						})
+					})
+				end
+
+				local parent = Instance.new("ScreenGui")
+				parent.Parent = game.CoreGui
+
+				local tree = createElement(ParentComponent)
+
+				local hostKey = "Some Key"
+				local instance = reconciler.mountVirtualNode(tree, parent, hostKey)
+
+				coroutine.resume(childCoroutine)
+
+				expect(#parent:GetChildren()).to.equal(1)
+
+				local frame = parent:GetChildren()[1]
+
+				expect(#frame:GetChildren()).to.equal(1)
+
+				reconciler.unmountVirtualNode(instance)
+			end)
+		end)
 	end)
 end


src/createReconciler.lua
```diff
@@ -66,13 +72,29 @@ local function createReconciler(renderer)
...
 		for childKey, childNode in pairs(virtualNode.children) do
 			local newElement = ElementUtils.getElementByKey(newChildElements, childKey)
 			local newNode = updateVirtualNode(childNode, newElement)
 
+			-- If updating this node has caused a component higher up the tree to re-render
+			-- and updateChildren to be re-entered for this virtualNode then
+			-- this result is invalid and needs to be disgarded.
+			if config.tempFixUpdateChildrenReEntrancy then
+				if virtualNode.updateChildrenCount ~= currentUpdateChildrenCount then
+					if newNode and not newNode ~= virtualNode.children[childKey] then
```
ZoteTheMighty, line 91: Looks like some double negatives here? Is this the same meaning, and is it clearer?
```suggestion
					if newNode ~= nil and newNode == virtualNode.children[childKey] then
```

src/createReconciler.lua
```diff
@@ -100,6 +122,18 @@ local function createReconciler(renderer)
 					virtualNode.legacyContext
 				)
 
+				-- If updating this node has caused a component higher up the tree to re-render
+				-- and updateChildren to be re-entered for this virtualNode then
+				-- this result is invalid and needs to be disgarded.
```
ZoteTheMighty, line 127: Small spelling fix
```suggestion
				-- this result is invalid and needs to be discarded.
```

src/createReconciler.lua
```diff
@@ -100,6 +122,18 @@ local function createReconciler(renderer)
...
+				if config.tempFixUpdateChildrenReEntrancy then
+					if virtualNode.updateChildrenCount ~= currentUpdateChildrenCount then
+						if childNode then
```
ZoteTheMighty, line 130: Confused about this... `childNode` is defined above on line 117, is it even possible for this condition to be false?

src/createReconciler.lua
```diff
@@ -100,6 +122,18 @@ local function createReconciler(renderer)
...
+							unmountVirtualNode(childNode)
+						end
+						return
```
ZoteTheMighty, line 133: It looks like what's happening here is that we're mounting a new child, then realizing we weren't supposed to do so and unmounting it. Is there a way we could avoid mounting it in the first place on line 117? Seems like we should move some of this logic up a bit.

src/createReconciler.lua
```diff
@@ -45,7 +45,13 @@ local function createReconciler(renderer)
 		local context = virtualNode.originalContext or virtualNode.context
 		local parentLegacyContext = virtualNode.parentLegacyContext
 
-		unmountVirtualNode(virtualNode)
+		if config.tempFixUpdateChildrenReEntrancy then
+			if not virtualNode.wasUnmounted then
+				unmountVirtualNode(virtualNode)
+			end
```
ZoteTheMighty, line 51: So I take this to mean that the re-entrancy was, in some cases, causing the same node to be unmounted more than once, is that the case? Can we get a comment here that explains the scenario that might trigger it, the way we have with the other blocks you added?

ZoteTheMighty (changes requested):
Some comments about the structure of the solution. Is the `wasUnmounted` logic a fix for something that was already happening? Or is it an additional solution to a _new_ bug that happens when the rest of this new logic is enabled?

src/createReconciler.lua
ConorGriffin37: The mounting of the child node is what causes it to no longer be needed as we re-renter updateChildren for the same node in the tree again.

src/createReconciler.lua
ConorGriffin37: mountVirtualNode can return nil if the element is a boolean

ConorGriffin37:
The wasUnmounted logic is a fix for something that could already happen. I was initially worried that double unmounting could be caused by this new logic (and still am) but it does actually already happen in the original bug.

Old way this could happen:
updateChildren happens for the first time
node is unmounted by call to updateVirtualNode() on line 84
due to event or callbacks triggered by this unmounting, another component higher up the tree starts re-rendering
updateChildren is renentered for the second time
node is unmounted again by another call to updateVirtualNode on line 84

src/createReconciler.lua
ConorGriffin37: Ah right, this actually should be just newNode ~= virtualNode.children[childKey]. I originally had it as 
not newNode == virtualNode.children[childKey] and must have messed up when changing it. I'll see can I add a test that would catch this issue.

src/createReconciler.lua
ConorGriffin37: Added a comment for this

src/createReconciler.lua
ConorGriffin37: Added a new test that catches this bug

src/RobloxRenderer.spec.lua
```diff
@@ -1025,5 +1025,316 @@ return function()
...
+
+				local ParentComponent = Component:extend("ParentComponent")
+
+				function ParentComponent:init()
+					self:setState({
+						count = 1
+					})
+
+					self.onDidMountCallback = function()
+						if self.state.count < 5 then
```
oltrep, line 1081: Is there a specific reason why `5`? To me it looks like `onDidMountCallback` should be called only once, so `self.state.count` should be `1` and then turn into `2`. If that is correct then maybe we could add an explicit assertion here that shows that (i.e. `expect(self.state.count).to.equal(1)`).

src/RobloxRenderer.spec.lua
```diff
@@ -1025,5 +1025,316 @@ return function()
...
+							self:setState({
```
oltrep, line 1209: Same thing as above ⬆️ (or below, I don't know how github will order my comments 😄 )

src/createReconciler.lua
oltrep: I wonder if this could be related to #235? Would it fix that issue?

oltrep (commented):
I am not super familiar with Roact's internals but the change looks good to me

ZoteTheMighty (approved):
This looks reasonably safe to me. Thanks for doing the legwork and pinning the changes with testing.

src/RobloxRenderer.spec.lua
ConorGriffin37: No specific reason why 5, just want to make sure there isn't a scenario where this test starts looping an unreasonable amount of times. Before the bug fix, onDidMountCallback would actually be called multiple times due to the duplicate components being mounted. I'll add the expect here

src/createReconciler.lua
ConorGriffin37: I don't think this would fix that issue, seems to be more about extra renders than extra unmounts.

src/RobloxRenderer.spec.lua
ConorGriffin37: Actually, onDidMountCallback is still called twice here. Basically, two different LowestComponent are still mounted so this is actually called twice. The first LowestComponent to be mounted is unmounted to prevent the duplication issue.

oltrep (approved):

PR #318, opened 2021-09-16
Update TestEZ and fix tests so they work in studio

oltrep:
TestEZ submodule has been updated to latest.

The usage of `CoreGui` was throwing permission errors when running tests within Studio so I've replaced it with a Folder instance in ReplicatedStorage. Since I've updated TestEZ, I was able to use `beforeEach` and `afterEach` to clean up that instance.

Checklist before submitting:
* [x] ~Added entry to `CHANGELOG.md`~
* [x] ~Added/updated documentation~

### DIFF
--- a/bin/run-tests.server.lua
+++ b/bin/run-tests.server.lua
@@ -12,14 +12,15 @@ Roact.setGlobalConfig({
 	["elementTracing"] = true,
 	["propValidation"] = true,
 })
-local results = TestEZ.TestBootstrap:run(ReplicatedStorage.Roact, TestEZ.Reporters.TextReporter)
+local results = TestEZ.TestBootstrap:run(
+	{ ReplicatedStorage.Roact },
+	TestEZ.Reporters.TextReporter
+)
 
-local statusCode = results.failureCount == 0 and 0 or 1
+local statusCode = (results.failureCount == 0 and #results.errors == 0) and 0 or 1
 
 if __LEMUR__ then
-	if results.failureCount > 0 then
-		os.exit(statusCode)
-	end
+	os.exit(statusCode)
 elseif isRobloxCli then
-	ProcessService:Exit(statusCode)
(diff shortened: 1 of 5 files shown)


bin/run-tests.server.lua
```diff
@@ -12,14 +12,15 @@ Roact.setGlobalConfig({
...
+local statusCode = (results.failureCount == 0 and #results.errors == 0) and 0 or 1
 
 if __LEMUR__ then
-	if results.failureCount > 0 then
-		os.exit(statusCode)
-	end
+	os.exit(statusCode)
 elseif isRobloxCli then
-	ProcessService:Exit(statusCode)
+	ProcessService:ExitAsync(statusCode)
```
hmallen99, line 25: What's the reason for this change? Just for my own learning.

place.project.json
```diff
@@ -11,7 +11,7 @@
       },
 
       "TestEZ": {
-        "$path": "modules/testez/lib"
+        "$path": "modules/testez"
```
hmallen99, line 14: Why does this not need to path to modules/testez/src, like in bin/spec.lua? Does it just automatically load the src?

hmallen99 (approved):
Looks good to me, all the tests still seem to be working. I had a couple questions, but nothing blocking

bin/run-tests.server.lua
oltrep: At some point the API just changed 😅 Blocking functions are usually marked with an `Async` suffix that's why it got corrected

place.project.json
oltrep: When you give a path to a directory to Rojo, it will look for a file named `default.project.json` and use that to map the directory to instances. If it does not exist, it'll default to just the regular file to instance mapping rules.

I think this file was probably written before Rojo was able to do that

PR #320, opened 2021-09-16
Fix bug in context consumers

### LINKED ISSUE #319: Error `Listeners can only be disconnected once` in context consumers
[PR #315](https://github.com/Roblox/roact/pull/315) uncovered that `willUnmount` can be called more than once in certain cases.

In some scenario, a context consumer component can have its `willUnmount` method called twice, which will cause this error:
```
ReplicatedStorage.Roact.createSignal:37: Listeners can only be disconnected once.
```

oltrep:
Closes #319.

PR #315 uncovered that willUnmount can be called more than once in certain cases. This PR adds a test that demonstrate the problem and fixes it.

Checklist before submitting:
* [x] ~Added/updated documentation~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased Changes
 
+* Fixed `Listeners can only be disconnected once` from context consumers. ([#320](https://github.com/Roblox/roact/pull/320))
+
 ## [1.4.1](https://github.com/Roblox/roact/releases/tag/v1.4.1) (August 12th, 2021)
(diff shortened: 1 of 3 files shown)


matthargett (approved):

ZoteTheMighty:
Is this problem related to something else going on with the component lifecycle, or is it unique to context?

oltrep:
> 
> 
> Is this problem related to something else going on with the component lifecycle, or is it unique to context?

The error comes specifically from context in this case, but the root cause is that `willUnmount` can be called twice on an element. The test case was adapted from a test from #315 which showed that behavior

ZoteTheMighty (approved):

PR #313, opened 2021-07-21
Silence indentation warning in script analysis

chasedig:
This commit silences the warning in Roblox Studio's "Script Analysis" Window

Error silenced:
W006: (285,3) Statement spans multiple lines; use indentation to silence

Script Analysis Warning:
RobloxStudioBeta_X2kVEYpQ0f

Script Analysis Warning Fixed:
RobloxStudioBeta_zPgptKUsGu

Checklist before submitting:
- [x] Added entry to `CHANGELOG.md`
- Added/updated relevant tests: Tests would not be possible as I do not believe you can interrogate/query the Script Analysis window. However I have attached an image of the script analysis window after the fix.
- Added/updated documentation: Update would not have affected anything in the documentation. Minimal change.

PR #321, opened 2021-09-17
Enforce code style and linting to eliminate Studio indentation warnings

matthargett:
Supercedes PR #313 . 

Rather than spot fix one styling issue, let's mistake-proof it.

Checklist before submitting:
CI-only and non-functional, so no CHANGELOG or test updates needed
```
```


.github/workflows/ci.yml
```diff
@@ -22,23 +22,41 @@ jobs:
       with:
         luaVersion: "5.1"
 
-    - uses: leafo/gh-actions-luarocks@v4
+    - name: get foreman and run foreman install
+      uses: rojo-rbx/setup-foreman@v1
+      with:
+        version: "^1.0.1"
+        token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
ZoteTheMighty, line 29: Remove this line and it should stop tripping over this step

selene.toml
```diff
@@ -0,0 +1,19 @@
+std = "roblox+testez"
+
+[config]
+empty_if = { comments_count = true }
+unused_variable = { ignore_pattern = "result|ok|^_" }
+# this comes up when translating nested try/finally scenarios
```
ZoteTheMighty, line 6: Looks like there's some settings and comments here from another project that likely don't apply; could you double check this file and clean those up? We may not need some of these particular settings to be non-default.

ZoteTheMighty (commented):
Can you doublecheck this in Studio and make sure this eliminates the warning, and doesn't introduce any others? StyLua and Selene may not share the same expectations as the script analyzer

matthargett:
> Can you doublecheck this in Studio and make sure this eliminates the warning, and doesn't introduce any others? StyLua and Selene may not share the same expectations as the script analyzer

Verified, and went a little further to pin a few files as strict/nonstrict.

ZoteTheMighty (approved):
Looks good!

ZoteTheMighty:
Took a closer look and it seems like you made changes to the `testez` submodule, any reason?

.github/workflows/ci.yml
```diff
@@ -24,21 +24,41 @@ jobs:
 
     - uses: leafo/gh-actions-luarocks@v4
 
+    - name: get foreman and run foreman install
+      uses: rojo-rbx/setup-foreman@v1
```
sasial-dev, line 28: Why are you using `rojo-rbx/setup-foreman@v1`
If I recall correctly, it's actually `Roblox/setup-foreman`

sasial-dev (changes requested):
Minor Pickup from me.

.github/workflows/ci.yml
matthargett: great question! the foreman project was spun out into the community where it could be more consistently maintained.

matthargett:
> Took a closer look and it seems like you made changes to the `testez` submodule, any reason?

Not sure how that happened. Let me try reverting that.

PR #324, opened 2021-10-01
Remove tempFixUpdateChildrenReEntrancy

ConorGriffin37:
We have had this turned on in a few projects for the last few months and it has proven sound. Remove tempFixUpdateChildrenReEntrancy so it is fixed by default.


ZoteTheMighty (approved):
This looks good to me, thanks for the cleanup

PR #372, opened 2022-06-09
Setup basic Luau analysis

ghostnaps:
This sets up the repo with Luau analysis and fixes the initial type errors

Checklist before submitting:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,11 @@ jobs:
         luarocks install luacov
         luarocks install luacov-reporter-lcov
 
+    - name: Install and run darklua
+      run: |
+        cargo install --git https://gitlab.com/seaofvoices/darklua.git#v0.7.0
+        darklua process src/ src/ --format retain-lines
+
     - name: Test
       run: |
         lua -lluacov bin/spec.lua
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,9 +71,7 @@ jobs:
         luarocks install luacov-reporter-lcov
 
     - name: Install and run darklua
-      run: |
-        cargo install --git https://gitlab.com/seaofvoices/darklua.git#v0.7.0
-        darklua process src/ src/ --format retain-lines
+      run: darklua process src/ src/ --format retain-lines
 
     - name: Test
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,24 @@ on:
     - master
 
 jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: Roblox/setup-foreman@v1
+        with:
+          version: "^1.0.1"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Download global Roblox types
+        shell: bash
+        run: curl -O https://raw.githubusercontent.com/JohnnyMorganz/luau-analyze-rojo/master/globalTypes.d.lua
+
+      - name: Analyze
+        shell: bash
+        run: luau-analyze --project=default.project.json --defs=globalTypes.d.lua --defs=testez.d.lua src/
+
   test:
     runs-on: ubuntu-latest
 
--- a/src/Binding.lua
+++ b/src/Binding.lua
@@ -134,7 +134,7 @@ function BindingInternalApi.join(upstreamBindings)
 				disconnect()
 			end
 
-			disconnects = nil
+			disconnects = nil :: any
 		end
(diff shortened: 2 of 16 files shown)


src/assertDeepEqual.lua
```diff
@@ -6,7 +6,7 @@
 	This should only be used in tests.
 ]]
 
-local function deepEqual(a, b)
+local function deepEqual(a, b): (boolean, string?)
```
matthargett, line 9: This should also work and more precisely express the scenarios:
```suggestion
local function deepEqual(a, b): (true, nil) | (false, string)
```

src/assertDeepEqual.lua
```diff
@@ -19,7 +19,7 @@ local function deepEqual(a, b)
 			visitedKeys[key] = true
 
 			local success, innerMessage = deepEqual(value, b[key])
-			if not success then
+			if not success and innerMessage then
```
matthargett, line 22: Unless there's a narrowing bug lingering, using the above type suggestion should eliminate the needs for these extra checks. If it doesn't, add a `-- FIXME Luau` comment and we can help file a Luau issue.

src/createReconciler.spec.lua
```diff
@@ -302,7 +302,9 @@ return function()
 			local node = noopReconciler.mountVirtualNode(emptyFragment, nil, "test")
 
 			expect(node).to.be.ok()
-			expect(next(node.children)).to.never.be.ok()
+
+			local nextNode = next(node.children)
```
matthargett, line 306: why was this extraction necessary? to force/avoid widening? if so, please mark with `-- FIXME Luau`.

src/strict.lua
```diff
@@ -1,4 +1,3 @@
---!nonstrict
```
matthargett, line 1: can you try changing these !nonstrict removals to !strict? if they're not strict-clean, leave the !nonstrict so folks who want to strictify checking on their app+packages don't get noise they can't do something about. (If they're impossible to make clean due to Luau issue, also add a `-- FIXME Luau`.

src/Binding.lua
```diff
@@ -134,7 +134,7 @@ function BindingInternalApi.join(upstreamBindings)
 				disconnect()
 			end
 
-			disconnects = nil
+			disconnects = nil :: any
```
matthargett, line 137: it would be better to make the delcaration of `disconnects` have an explicit type that includes nil (eg `local disconnects: Array<any> | nil = {}`). (Once deferred constraint resolution and some other Luau features land, it will not longer greedily assume the first assignment it sees is the only type possibility for the whole closure.)

src/Component.lua
```diff
@@ -341,7 +341,7 @@ function Component:__mount(reconciler, virtualNode)
 		instance:didMount()
 	end
 
-	if internalData.pendingState ~= nil then
+	if (internalData :: any).pendingState ~= nil then
```
matthargett, line 344: I'd suggest adding `pendingState = nil` to the `internalData` table declaration instead of this hard cast operator.

.github/workflows/ci.yml
```diff
@@ -52,6 +70,11 @@ jobs:
         luarocks install luacov
         luarocks install luacov-reporter-lcov
 
+    - name: Install and run darklua
```
matthargett, line 73: we should be able to install darklua via foreman as of January, thanks to @jeparlefrancais adding gitlab support to foreman :)

matthargett (commented):
Great work! Some minor suggestions.

src/createReconciler.spec.lua
ZoteTheMighty: Probably because `next` returns multiple values and `expect` only accepts 1

matthargett: So, a Luau issue then? I'll ask in the slack

src/assertDeepEqual.lua
ghostnaps: Had no idea we could use boolean literals. Awesome!

src/Binding.lua
ghostnaps: This leads to issues with indexing the `disconnects` variable since Luau now thinks it could be nil at any point. I think it would be a bit overkill to check `if disconnects then` so maybe we should just stick with casting to any

Screen Shot 2022-06-10 at 11 41 59 AM

ghostnaps:
Got a few analysis errors left from enabling strict mode in createReconsiler. My Luau knowledge is run up against a wall here so I would appreciate somebody taking a pass to clean up these remaining issues

ZoteTheMighty (approved):

ZoteTheMighty:
This is a good start towards type strictness! Thanks for introducing the tooling as well.

PR #381, opened 2023-01-04
Pass "Type" and "ElementKind" to init.lua table

bark-ing:
This is more of a personal request but I believe it's providing a use case for `RoactType` and `ElementKind`.  These are used as keys within Elements and Fragments and not accessible by string. Just a thought that it should be accessible through `Roact` module and take up less space to include it in a module than its own variable.

### DIFF
--- a/src/init.lua
+++ b/src/init.lua
@@ -32,6 +32,9 @@ local Roact = strict({
 	Event = require(script.PropMarkers.Event),
 	Ref = require(script.PropMarkers.Ref),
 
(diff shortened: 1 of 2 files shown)


foxesbark:
* Added `Type` and `ElementKind` to `init.lua`. ([#381](https://github.com/Roblox/roact/pull/381))

foxesbark:
I have read the CLA Document and I hereby sign the CLA

PR #387, opened 2023-06-17
Remove trimming from element tracing

Ew-Developer:
Original code said there was no to specify a trace level without also specifying a message, but you can just set the message to nil to achieve this.

This change makes it so we don't have to trim the traceback to remove the newline.

*Put your pull request body here!*

Checklist before submitting:


Awokein (approved):
Simple but necessary change. Good job 👍

